PR #43, opened 2020-05-06
Tickets/dm 24590 2

teweitsai:


python/lsst/ts/wep/deblend/DeblendDefault.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Raises
+        ------
+        NotImplementedError
+            Child class should implement this.
+        """
+        raise NotImplementedError("Child class should implement this.")
+
+
+if __name__ == "__main__":
```
pkubanek: salobj isn't using this, and I haven't seen that at any Python style guide as far as I remember. I don't see reason for adding those two lines, as IMHO they don't bring any functionality - I will ask for them to be removed unless there is some reasons for their existence I am missing

python/lsst/ts/wep/deblend/DeblendDonutFactory.py
```diff
@@ -0,0 +1,36 @@
...
+            The deblend donut type is not supported.
+        """
+
+        if (deblendDonutType == DeblendDonutType.Adapt):
+            return DeblendAdapt()
+        else:
+            raise ValueError("The %s is not supported." % deblendDonutType)
+
+
+if __name__ == "__main__":
```
pkubanek: See previous comment.

pkubanek (approved):
See comments for __main__ / pass. Otherwise looks fine.

python/lsst/ts/wep/deblend/DeblendDefault.py
teweitsai: Done in commit: 2ea6a7c. BTW, I tried to let `ts_salobj` only couples with `ts_MTAOS`. `ts_wep` will be in the DM scientific pipeline in the final and will not have any relationship with `ts_salobj`, which is specific to T&S team. Thanks!

python/lsst/ts/wep/deblend/DeblendDonutFactory.py
teweitsai: Done in commit: 2ea6a7c. Thanks!

PR #45, opened 2020-05-18
Improved handling of IO errors.

pkubanek:
The original code doesn't handled un-readable files and other corner cases.
Catching IOError handles more problems.


teweitsai (approved):
This update looks good to me! Please help to update the [versionHistory](
https://github.com/lsst-ts/ts_wep/blob/tickets/DM-24610/doc/versionHistory.rst) for this update. Thanks!

PR #47, opened 2020-06-01
Tickets/dm 25046 2

teweitsai:


include/lsst/ts/wep/cwfs/MathCwfs.h
```diff
@@ -0,0 +1,125 @@
...
+ * @param[in] arrayX  X coordinate on pupil plane
+ * @param[in] arrayY  Y coordinate on pupil plane
+ * @param[in] axis  Direction of gradient ("dx" or "dy")
+ * @return the corrected parameters for off-axis distortion
+ */
+py::array_t<double> poly10Grad(py::array_t<double> arrayC,
+                               py::array_t<double> arrayX,
+                               py::array_t<double> arrayY, std::string axis);
+
+PYBIND11_MODULE(mathcwfs, m) {
```
r-owen: I suggest you use the same name here as the file -- they presently have different capitalization. Most python modules use lowercase, so perhaps name this file "mathcwfs.h"?

and similarly for the .cc file.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+
+        Rnl = sqrt_14 * num13 * r6;
+        temp += Z[26] * Rnl * s6;
+        temp += Z[27] * Rnl * c6;
+
+        S[ii] = temp;
+    }
+
+    return result;
+}
```
r-owen: Are these functions likely to be also wanted for C++ code? If so, please consider separating this and the header file into three files: a header and implementation for the C++ (with no pybind11 in them), plus a simple pybind11 wrapper.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+            y4 = y2 * y2;
+
+            // to make d an array with the same size as x
+            temp = Z[0] * 0 * x_c;
+            temp += Z[1] * 0;
+            temp += Z[2] * 0;
+
+            temp += Z[3] * sqrt_3 * 8 / den1;
+            temp += Z[4] * sqrt_6 * 0;
+            temp += Z[5] * sqrt_6 * 0;
```
r-owen: Please consider removing the lines that do not change the value of `temp` (which is most of these lines -- due to the right hand side having a multiplication by 0). I think the result would be easier to read.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+                    ((e4 - 5 * e2 * x2 - 5 * e2 * y2 + 3 * e2 + 5 * x4 +
+                      10 * x2 * y2 - 5 * x2 + 5 * y4 - 5 * y2 + 1) *
+                     (e4 - 15 * e2 * x2 - 15 * e2 * y2 + 3 * e2 + 25 * x4 +
+                      50 * x2 * y2 - 15 * x2 + 25 * y4 - 15 * y2 + 1)) /
+                    den2_11;
+
+            out[ii] = temp;
+        }
+    }
+    return result;
```
r-owen: Please raise an exception if `atype` is neither "1st" nor "2nd".

Please also consider changing `atype` to an enum class (though that is a breaking change and I don't know how widely this code is used). The argument is that it makes the API a bit clearer (the options are self-documenting) and catches errors a bit earlier.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+                    num10 / den10;
+
+            temp += Z[19] * sqrt_12 * 20 * y_c * (y2 - 3 * x2) / den11;
+            temp += Z[20] * sqrt_12 * 20 * x_c * (x2 - 3 * y2) / den11;
+
+            temp += Z[21] * sqrt_7 * 240 * xy * (2 * r2 - 1 - e2) / den12;
+
+            d[ii] = temp;
+        }
+    }
```
r-owen: Please raise an exception if axis is none of the above.

This is another case where making `axis` an enum class is worth considering.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+    auto result = py::array_t<double>(n);
+    ArrayNpInfo infoResult = getNpArrayInfo(result);
+
+    // Assign the variables
+    double *Z = infoZk.buf;
+    double *x = infoX.buf;
+    double *y = infoY.buf;
+    double *d = infoResult.buf;
+
+    // Parameters of constant
```
r-owen: For all the below values that are actually constant (which looks like quite a few of them) please declare them so, e.g.:
```
double const e2 = pow(e, 2);
```
It clarifies the intent of the code and offers the compiler an opportunity for optimization.

src/MathCwfs.cc
```diff
@@ -0,0 +1,1156 @@
...
+        for (size_t ii = 0; ii < n; ii++) {
+            x_c = x[ii];
+            y_c = y[ii];
+
+            x2 = x_c * x_c;
+            y2 = y_c * y_c;
+            x4 = x2 * x2;
+            y4 = y2 * y2;
+            xy = x_c * y_c;
+            r2 = x2 + y2;
```
r-owen: A gentle suggestion to consider allocating these as they are defined, and using `const` where appropriate.

I am on the fence about this. Generally it is considered best practice to declare variables when they are set and, of course, to use `const` where appropriate. But the code may be easier to read without the many repetitions of `double const`.

.githooks/pre-commit
```diff
@@ -1,3 +1,5 @@
 #!/bin/sh
 echo "Formatting Python code with black"
 black .
+echo "Formatting C++ code with clang-format"
+find . -name '*.cc' -o -name '*.h' | xargs clang-format -i --style="{IndentWidth: 4}"
```
r-owen, line 5: Nice!

tests/bsc/test_localDatabaseForStarFile.py
```diff
@@ -118,7 +118,8 @@ def testInsertDataByFileWithoutStar(self):
         self._createTable()
 
         skyFilePath = self._writeStarFile("noStar.txt")
-        idAll = self._insertDataToDbAndGetAllId(skyFilePath)
+        with self.assertWarns(UserWarning):
```
r-owen, line 121: Why the warning?

r-owen (changes requested):
Overall this looks excellent.

My primary requested change is that your functions that take a "what operation to perform" parameter should raise an exception if the value is not recognized -- and add a unit test for it. (A related change to consider is to make these parameters enum classes. That makes it much less likely that the value *can* be a non-supported value, though it may still be worth checking.)

src/MathCwfs.cc
teweitsai: I am not sure I understand the meaning. The functions (in `.cc` file) need the data type of `pybind11::array_t` and therefore, the `pybind11` is needed. Or your meaning is that I should put the code of `PYBIND11_MODULE` block in header file (`MathCwfs.h`) into another header file? Please help to clarify this. Thanks!

include/lsst/ts/wep/cwfs/MathCwfs.h
teweitsai: Done in the commit: 6d31edc. Thanks!

src/MathCwfs.cc
teweitsai: Updated in commit: 6d31edc. Thanks!

src/MathCwfs.cc
teweitsai: Updated in commit: 6d31edc. Thanks!

tests/bsc/test_localDatabaseForStarFile.py
teweitsai: `Numpy` will throw the `UserWarning` if load a text file without the content (only the comment `#` exists).

src/MathCwfs.cc
teweitsai: Usually, I only use the `const` in the function inputs to tell the user this specific input argument is a constant. I felt it is annoying to see a lot of `const` in the function realization, especially for the scientific calculation. I agree it is good practice to declare the `const`, but it may have some affection of readability.

src/MathCwfs.cc
r-owen: That is a good point and I should have seen that. A generic C++ version would have to use some other sequence type, presumably std::array. pybind11 can wrap all the usual C++ sequence types just fine.

But if you don't need it to be generic then there's no point worrying about it.

src/MathCwfs.cc
teweitsai: This implementation is based on the user manual of `pybind11`:
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html

It looks like this might be because of the buffer protocol, but I do not know the details. You could see the following as the example  in the `pybind11` manual:
```
#include <pybind11/pybind11.h>
#include <pybind11/numpy.h>

namespace py = pybind11;

py::array_t<double> add_arrays(py::array_t<double> input1, py::array_t<double> input2) {
    py::buffer_info buf1 = input1.request(), buf2 = input2.request();

    if (buf1.ndim != 1 || buf2.ndim != 1)
        throw std::runtime_error("Number of dimensions must be one");

    if (buf1.size != buf2.size)
        throw std::runtime_error("Input shapes must match");

    /* No pointer is passed, so NumPy will allocate the buffer */
    auto result = py::array_t<double>(buf1.size);

    py::buffer_info buf3 = result.request();

    double *ptr1 = (double *) buf1.ptr,
           *ptr2 = (double *) buf2.ptr,
           *ptr3 = (double *) buf3.ptr;

    for (size_t idx = 0; idx < buf1.shape[0]; idx++)
        ptr3[idx] = ptr1[idx] + ptr2[idx];

    return result;
}

PYBIND11_MODULE(test, m) {
    m.def("add_arrays", &add_arrays, "Add two NumPy arrays");
}
```

src/MathCwfs.cc
teweitsai: Done in the commit: 6d31edc. I am considering to put the `enum` in the Python level. But I am not sure. It will be another JIRA ticket after I have a decision. Thanks!

src/MathCwfs.cc
teweitsai: I understand your meaning now. `pybind11` supports the `list` and `array` types in C++. It is around 2017 (based on the history of `pybind11/numpy.h`), `pybind11` began to support the `numpy.ndarray` directly. I read the article to say the bottom implementation of `pybind11` for `numpy` is pretty similar to `numpy` itself's realization. Therefore, I think I could use the `pybind11/numpy.h` directly because I do not need to let the functions to be so general. They are just for some specific calculation to improve the performance. Thanks!

src/MathCwfs.cc
r-owen: I am a bit confused. Perhaps you are talking about the buffer pointers above my comment? I mean the code that follows my comment -- that you have noted as
```
// Parameters of constant
```
I don't think those have anything to do with pybind11.

src/MathCwfs.cc
r-owen: Thanks. pybind11 does a good job wrapping C++ enum classes. If you do eventually change the code I think you'll have to define the new C++ enum classes, wrap them with pybind11 and then all this should work fine from Python.

r-owen (approved):
Thank you for raising an exception for invalid values of those two string parameters, and adding a unit test.

This looks good. My one small suggestion is to use assertRaises as a context manager, but do as you like. Your tests work fine as written.

PR #48, opened 2020-09-12
adding doOverscan ISR setting as default

suberlak:


teweitsai (approved):
The update looks good to me. Great!

Issue #52, opened 2020-11-20
Remove the sims dependencies from WcsSol.py


jbkalmbach:
Switch them out to use `obs_lsst`.

teweitsai:
This issue needs to consider the rotation of WCS in `obs_lsst` carefully. This is related to #35.

jbkalmbach:
Closed by #54

PR #53, opened 2020-12-04
Updated to work with latest weekly.

tribeiro:
This mainly involves replacing the use of the deprecated package `obs_lsstSim` by the official camera geometry `from lsst.obs.lsst import LsstCamMapper`.

There are some issues when it comes to the test data, which will need to be regenerated with the correct mapping. I am skipping those tests so it will work for now but the right thing to do would be to regenerate the data. 

I am not entirely sure it is worth the trouble of merging this PR, we can discuss in the review process. In the meantime, I'll use this to build the MTAOS CSC for deployment with xml 7 and the latest weekly.


teweitsai:
Basically, this update conflicts with #52. The problem is not the test image but the coordinate system difference between `obs_lsst` and PhoSim (or `sims` package). The correct way is to wait until the #52 is done. We could talk before I go ahead to fail this PR. Thanks!

In addition, this PR will break the AOS simulation in UW team directly.

Some message left by Scott: lsst/sims_coordUtils@832d909. Based on his suggestion, I tried to use the `PhosimMapper` instead. But it looks like the geometry of `PhosimMapper` does not follow `LsstSimMapper` as well. The test is at [feature/w_2020_49](https://github.com/lsst-ts/ts_wep/tree/feature/w_2020_49) branch.

teweitsai:
The dependency of `sims` package had been removed in commit: 5c5288e.

PR #58, opened 2021-01-07
Adding new centroid finding algorithm: CentroidConvolveTemplate.

jbkalmbach:
Hi @teweitsai 

This is to start merging code developed in the UW branch into the main codebase in pieces that make sense. Here is an update to add in the Template Convolution centroiding method we created.

### DIFF
--- a/doc/uml/cwfsClass.uml
+++ b/doc/uml/cwfsClass.uml
@@ -9,7 +9,7 @@ CentroidDefault <|-- CentroidConvolveTemplate
 CentroidFindFactory ..> CentroidRandomWalk
 CentroidFindFactory ..> CentroidOtsu
 CentroidFindFactory ..> CentroidConvolveTemplate
-CentroidRandomWalk ..> CentroidConvolveTemplate
+CentroidConvolveTemplate *-- CentroidRandomWalk
 Image ..> CentroidFindFactory
 Image *-- CentroidDefault
 BaseCwfsTestCase ..> CompensableImage
--- a/doc/uml/cwfsClass.uml
+++ b/doc/uml/cwfsClass.uml
@@ -5,8 +5,11 @@ Algorithm -- CompensableImage
 CompensableImage ..> Instrument
 CentroidDefault <|-- CentroidRandomWalk
 CentroidDefault <|-- CentroidOtsu
+CentroidDefault <|-- CentroidConvolveTemplate
 CentroidFindFactory ..> CentroidRandomWalk
 CentroidFindFactory ..> CentroidOtsu
+CentroidFindFactory ..> CentroidConvolveTemplate
+CentroidRandomWalk ..> CentroidConvolveTemplate
 Image ..> CentroidFindFactory
 Image *-- CentroidDefault
 BaseCwfsTestCase ..> CompensableImage
--- a/python/lsst/ts/wep/Utility.py
+++ b/python/lsst/ts/wep/Utility.py
@@ -360,7 +360,7 @@ def getCentroidFindType(centroidFindType):
     Parameters
     ----------
     centroidFindType : str
-        Centroid find algorithm to use (randomWalk or otsu).
+        Centroid find algorithm to use (randomWalk, otsu, or convolveTemplate).
 
     Returns
     -------
--- a/python/lsst/ts/wep/Utility.py
+++ b/python/lsst/ts/wep/Utility.py
@@ -62,6 +62,7 @@ class ImageType(IntEnum):
 class CentroidFindType(IntEnum):
     RandomWalk = 1
     Otsu = auto()
+    ConvolveTemplate = auto()
 
 
 class DeblendDonutType(IntEnum):
@@ -376,6 +377,8 @@ def getCentroidFindType(centroidFindType):
         return CentroidFindType.RandomWalk
     elif centroidFindType == "otsu":
         return CentroidFindType.Otsu
+    elif centroidFindType == "convolveTemplate":
+        return CentroidFindType.ConvolveTemplate
     else:
         raise ValueError("The %s is not supported." % centroidFindType)
 
--- a/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
+++ b/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
@@ -1,0 +1,49 @@
+import numpy as np
+
+from lsst.ts.wep.cwfs.CentroidRandomWalk import CentroidRandomWalk
+from scipy.signal import correlate
+from sklearn.cluster import KMeans
+
+
+class CentroidConvolveTemplate(CentroidRandomWalk):
+
+    """CentroidDefault child class to get the centroid of donut by
+    convolution."""
+
+    def getCenterFromTemplateConv(self, imageBinary, templateImgBinary,
+                                  n_donuts):
+        """Get the centers of the donuts in the binary image.
+
+        Parameters
+        ----------
+        imageBinary: numpy.ndarray
+            Binary image of postage stamp
+
+        templateImgBinary: numpy.ndarray
+            Binary image of template donut
+
+        Returns
+        -------
+        nx, ny: float
+            x, y pixel coordinates for donut centroid
+        """
+
+        temp_convolve = correlate(imageBinary, templateImgBinary, mode='same', method='fft')
+        ranked_convolve = np.argsort(temp_convolve.flatten())[::-1]
+        cutoff = len(np.where(temp_convolve.flatten() > 0.95*np.max(temp_convolve))[0])
+        ranked_convolve = ranked_convolve[:cutoff]
+        nx, ny = np.unravel_index(ranked_convolve, np.shape(imageBinary))
+
+        kmeans = KMeans(n_clusters=n_donuts).fit(np.array([nx, ny]).T)
+        labels = kmeans.labels_
+
+        cent_x = []
+        cent_y = []
+
+        for label_num in range(n_donuts):
+            nx_label, ny_label = np.unravel_index(ranked_convolve[labels == label_num][0],
+                                                  np.shape(imageBinary))
+            cent_x.append(nx_label)
+            cent_y.append(ny_label)
+
+        return cent_x, cent_y
--- a/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
+++ b/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
@@ -32,8 +32,7 @@ class CentroidConvolveTemplate(CentroidDefault):
     """CentroidDefault child class to get the centroid of donut by
     convolution."""
 
-    def getCenterAndR(self, imgDonut, templateDonut, nDonuts,
-                      peakThreshold=0.95):
+    def getCenterAndR(self, imgDonut, templateDonut, nDonuts, peakThreshold=0.95):
         """Get the centroid data and effective weighting radius.
 
         Parameters
@@ -65,12 +64,13 @@ class CentroidConvolveTemplate(CentroidDefault):
         imgBinary = self._centRandomWalk.getImgBinary(imgDonut)
         templateBinary = self._centRandomWalk.getImgBinary(templateDonut)
 
-        return self.getCenterAndRfromImgBinary(imgBinary, templateBinary,
-                                               nDonuts,
-                                               peakThreshold=peakThreshold)
+        return self.getCenterAndRfromImgBinary(
+            imgBinary, templateBinary, nDonuts, peakThreshold=peakThreshold
+        )
 
-    def getCenterAndRfromImgBinary(self, imgBinary, templateBinary, nDonuts,
-                                   peakThreshold=0.95):
+    def getCenterAndRfromImgBinary(
+        self, imgBinary, templateBinary, nDonuts, peakThreshold=0.95
+    ):
         """Get the centroid data and effective weighting radius.
 
         Parameters
@@ -98,15 +98,16 @@ class CentroidConvolveTemplate(CentroidDefault):
             Effective weighting radius.
         """
 
-        x, y = self.getCenterFromTemplateConv(imgBinary, templateBinary,
-                                              nDonuts,
-                                              peakThreshold=peakThreshold)
+        x, y = self.getCenterFromTemplateConv(
+            imgBinary, templateBinary, nDonuts, peakThreshold=peakThreshold
+        )
         radius = np.sqrt(np.sum(templateBinary) / np.pi)
 
         return x, y, radius
 
-    def getCenterFromTemplateConv(self, imageBinary, templateImgBinary,
-                                  nDonuts, peakThreshold=0.95):
+    def getCenterFromTemplateConv(
+        self, imageBinary, templateImgBinary, nDonuts, peakThreshold=0.95
+    ):
         """
         Get the centers of the donuts by convolving a binary template image
         with the binary image of the donut or donuts.
@@ -141,17 +142,13 @@ class CentroidConvolveTemplate(CentroidDefault):
 
         # We set the mode to be "same" because we need to return the same
         # size image to the code.
-        tempConvolve = correlate(
-            imageBinary, templateImgBinary, mode="same"
-        )
+        tempConvolve = correlate(imageBinary, templateImgBinary, mode="same")
 
         # Then we rank the pixel values keeping only those above
         # some fraction of the highest value.
         rankedConvolve = np.argsort(tempConvolve.flatten())[::-1]
         cutoff = len(
-            np.where(
-                tempConvolve.flatten() > peakThreshold * np.max(tempConvolve)
-            )[0]
+            np.where(tempConvolve.flatten() > peakThreshold * np.max(tempConvolve))[0]
         )
         rankedConvolve = rankedConvolve[:cutoff]
         nx, ny = np.unravel_index(rankedConvolve, np.shape(imageBinary))
--- a/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
+++ b/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
@@ -81,7 +81,7 @@ class CentroidConvolveTemplate(CentroidDefault):
         imgBinary = self.getImgBinary(imgDonut)
 
         if templateDonut is None:
-            templateBinary = templateDonut
+            templateBinary = copy(imgBinary)
         else:
             templateBinary = self.getImgBinary(templateDonut)
 
--- a/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
+++ b/python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
@@ -20,7 +20,7 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import numpy as np
-
+from copy import deepcopy
 from lsst.ts.wep.cwfs.CentroidDefault import CentroidDefault
 from lsst.ts.wep.cwfs.CentroidRandomWalk import CentroidRandomWalk
 from scipy.signal import correlate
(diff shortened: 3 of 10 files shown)


teweitsai (changes requested):
Please base on today's discussion to do the related code modification. Thanks!

doc/uml/cwfsClass.uml
```diff
@@ -5,11 +5,14 @@ Algorithm -- CompensableImage
 CompensableImage ..> Instrument
 CentroidDefault <|-- CentroidRandomWalk
 CentroidDefault <|-- CentroidOtsu
+CentroidDefault <|-- CentroidConvolveTemplate
 CentroidFindFactory ..> CentroidRandomWalk
 CentroidFindFactory ..> CentroidOtsu
+CentroidFindFactory ..> CentroidConvolveTemplate
+CentroidRandomWalk ..> CentroidConvolveTemplate
```
teweitsai: Based on your code, the relationship should be `CentroidConvolveTemplate *-- CentroidRandomWalk`. Please check the difference between the `dependency` and `composition` in UML class diagram. You may want to add the description of **CentroidConvolveTemplate** in `ts_wep/doc/content.rst`.

doc/uml/cwfsClass.uml
```diff
@@ -5,11 +5,14 @@ Algorithm -- CompensableImage
...
 Image ..> CentroidFindFactory
 Image *-- CentroidDefault
 BaseCwfsTestCase ..> CompensableImage
 BaseCwfsTestCase ..> Instrument
 BaseCwfsTestCase ..> Algorithm
-@enduml
+@enduml
```
teweitsai: Please add a new line.

python/lsst/ts/wep/Utility.py
```diff
@@ -376,6 +377,8 @@ def getCentroidFindType(centroidFindType):
         return CentroidFindType.RandomWalk
     elif centroidFindType == "otsu":
         return CentroidFindType.Otsu
+    elif centroidFindType == "convolveTemplate":
```
teweitsai, line 380: Please update the doc string of `getCentroidFindType()` as well. e.g. `Centroid find algorithm to use (randomWalk,  otsu, or convolveTemplate).`

python/lsst/ts/wep/cwfs/CentroidDefault.py
```diff
@@ -26,7 +26,7 @@
 class CentroidDefault(object):
     """Default Centroid class."""
 
-    def getCenterAndR(self, imgDonut):
+    def getCenterAndR(self, imgDonut, **kwargs):
```
teweitsai, line 29: Please add the description of "**kwargs" in the function doc string.

python/lsst/ts/wep/cwfs/CentroidDefault.py
```diff
@@ -48,7 +48,7 @@ def getCenterAndR(self, imgDonut):
 
         return self.getCenterAndRfromImgBinary(imgBinary)
 
-    def getCenterAndRfromImgBinary(self, imgBinary):
+    def getCenterAndRfromImgBinary(self, imgBinary, **kwargs):
```
teweitsai, line 53: Please add the description of "**kwargs" in the function doc string.

python/lsst/ts/wep/cwfs/CentroidFindFactory.py
```diff
@@ -39,7 +40,7 @@ def createCentroidFind(centroidFindType):
 
         Returns
         -------
-        CentroidRandomWalk, CentroidOtsu
+        CentroidRandomWalk, CentroidOtsu, CentroidConvolveTemplate
```
teweitsai: You could just say the child class of `CentroidDefault` to avoid the future's modification if there was a new child class added at that time.

python/lsst/ts/wep/cwfs/CentroidFindFactory.py
```diff
@@ -52,5 +53,7 @@ def createCentroidFind(centroidFindType):
             return CentroidRandomWalk()
         elif centroidFindType == CentroidFindType.Otsu:
             return CentroidOtsu()
+        elif centroidFindType == CentroidFindType.ConvolveTemplate:
+            return CentroidConvolveTemplate()
```
teweitsai, line 57: Please add the related test case in `ts_wep/tests/cwfs/test_centroidFindFactory.py`.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+from lsst.ts.wep.cwfs.CentroidDefault import CentroidDefault
+from lsst.ts.wep.cwfs.CentroidRandomWalk import CentroidRandomWalk
+from scipy.signal import correlate
+from sklearn.cluster import KMeans
+
+
+class CentroidConvolveTemplate(CentroidDefault):
+
+    """CentroidDefault child class to get the centroid of donut by
+    convolution."""
```
teweitsai: Maybe "... by convolution with the template image."

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+    def getCenterAndR(
+        self, imgDonut, templateDonut=None, nDonuts=1, peakThreshold=0.95
+    ):
+        """Get the centroid data and effective weighting radius.
+
+        Parameters
+        ----------
+        imgDonut : numpy.ndarray
+            Donut image.
+        templateDonut : None or numpy.ndarray
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Template image for a single donut. If set to None
+            then the image will be convolved with itself. (The Default is None)
+        nDonuts : int
```
teweitsai: The argument is "optional". In addition, this function will not return the `float` but `list` if the `nDonuts > 1`. This will be inconsistent with doc string for the return value. I will suggest to remove this argument and put the `nDonut=1` when you call the `getCenterFromTemplateConv()` from this function or `getCenterAndRfromImgBinary()`. In addition, did you notice that your return type is `list` instead of `float` for x and y?

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Number of donuts in the image. (The Default is 1)
+        peakThreshold : float
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        float
+            Centroid x.
+        float
+            Centroid y.
+        float
+            Effective weighting radius.
+        """
+
+        if templateDonut is None:
+            templateDonut = deepcopy(imgDonut)
```
teweitsai: `deepcopy()` is pretty expensive. I think `copy.copy()` should be good enough for a simple `numpy.array`. The idea is to make sure `imgDonut` and `templateDonut` are not the same object, which can use `id(imgDonut)` and `id(templateDonut)` to check. BTW, do you need this here? It looks like `getCenterFromTemplateConv()` will do the `deepcopy()` as well?

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self._centRandomWalk = CentroidRandomWalk()
```
teweitsai: Please put the declaration of new class attribute to `__init__()`. The developers are used to read the initialization function to see the available attributes for a class. In addition, you could add the `super()` to this to avoid the future's modification if there was the new attributes added for `CentroidDefault` in the future.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+    def getCenterAndRfromImgBinary(
+        self, imgBinary, templateBinary=None, nDonuts=1, peakThreshold=0.95
+    ):
+        """Get the centroid data and effective weighting radius.
+
+        Parameters
+        ----------
+        imgBinary : numpy.ndarray
+            Binary image of donut.
+        templateBinary : None or numpy.ndarray
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Binary image of template for a single donut. If set to None
+            then the image will be convolved with itself. (The Default is None)
+        nDonuts : int
```
teweitsai: I will suggest to remove this argument as explained in the above. The main reason is the returned data type.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Number of donuts in the image. (The Default is 1)
+        peakThreshold : float
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        float
+            Centroid x.
+        float
+            Centroid y.
+        float
+            Effective weighting radius.
+        """
+
+        if templateBinary is None:
+            templateBinary = deepcopy(imgBinary)
```
teweitsai: Maybe `copy.copy()` is good enough. Please see the above comments. BTW, do you need this here? It looks like `getCenterFromTemplateConv()` will do the `deepcopy()` as well?

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+
+        x, y = self.getCenterFromTemplateConv(
+            imgBinary,
+            templateImgBinary=templateBinary,
+            nDonuts=nDonuts,
```
teweitsai: Please use `nDonuts=1` here instead. The reason is explained above to make sure the consistent return data type as other child class of **CentroidDefault**.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            peakThreshold=peakThreshold,
+        )
+        radius = np.sqrt(np.sum(templateBinary) / np.pi)
+
+        return x, y, radius
```
teweitsai: You may need to do `return x[0], y[0], radius` instead. The return type of `getCenterFromTemplateConv()` is `list`.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        Peaks will appear as bright spots in the convolved image. Since we
+        use binary images the brightness of the stars does not matter and
+        the peaks of any stars in the image should have about the same
+        brightness if the template is correct.
+
+        Parameters
+        ----------
+        imageBinary: numpy.ndarray
+            Binary image of postage stamp.
+        templateImgBinary: None or numpy.ndarray
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Binary image of template donut. If set to None then the image
+            is convolved with itself. (The default is None)
+        nDonuts: int
```
teweitsai: This argument is "optional". What will happen if people put `nDonuts` to be 0 or negative value? Maybe you could check the input value and raise the related error?

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+            Number of donuts there should be in the binary image. (The default
+            is 1)
+        peakThreshold: float
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        Returns
+        -------
+        list
+            X pixel coordinates for donut centroid.
+        list
+            Y pixel coordinates for donut centroid.
+        """
+
+        if templateImgBinary is None:
+            templateImgBinary = deepcopy(imageBinary)
```
teweitsai: I think maybe the `copy.copy()` is good enough as explained. Please help to double check. Thanks!

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        # size image to the code.
+        tempConvolve = correlate(imageBinary, templateImgBinary, mode="same")
+
+        # Then we rank the pixel values keeping only those above
+        # some fraction of the highest value.
+        rankedConvolve = np.argsort(tempConvolve.flatten())[::-1]
+        cutoff = len(
+            np.where(tempConvolve.flatten() > peakThreshold * np.max(tempConvolve))[0]
+        )
+        rankedConvolve = rankedConvolve[:cutoff]
```
teweitsai: It is better to use a new argument such as `rankedConvolveCutoff`. This is to consider the future's maintenance if someone wants to check the original value of `rankedConvolve` before the cutoff for the debug or reuse.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+        nx, ny = np.unravel_index(rankedConvolve, np.shape(imageBinary))
+
+        # Then to find peaks in the image we use K-Means with the
+        # specified number of donuts
+        kmeans = KMeans(n_clusters=nDonuts).fit(np.array([nx, ny]).T)
+        labels = kmeans.labels_
```
teweitsai: It looks like you are using the internal attribute of `KMeans`. Does it provide any function API such as the `get()` function to get the value you need? The benefit to use the `get()` function is because it is less likely to be renamed compared with the internal attribute.

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -0,0 +1,193 @@
...
+
+        centX = []
+        centY = []
```
teweitsai, line 195: You could put this two lines just before `for labelNum in range(nDonuts):` to improve the readability.
```
centX = []
centY = []
for labelNum in range(nDonuts):
```

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -0,0 +1,131 @@
...
+        self.centroidConv = CentroidConvolveTemplate()
+
+    def _createData(self):
+
+        radiusInner = 20
+        radiusOuter = 40
+
+        # Create two images. One with a single donut and one with two donuts.
+        singleDonut = np.zeros((160, 160))
+        doubleDonut = np.zeros((160, 160))
```
teweitsai: You could do the following for the test in future and improve the reusability:
```
def _createData(self, radiusInner, radiusOuter, imageSize):

    # Create two images. One with a single donut and one with two donuts.
    singleDonut = np.zeros((imageSize, imageSize))
    doubleDonut = np.zeros((imageSize, imageSize))
```
You may need to update the following code as well to reduce the number of hard-coded values.

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -0,0 +1,131 @@
...
+        return singleDonut, doubleDonut
+
+    def testGetCenterAndR(self):
+
+        singleDonut, doubleDonut = self._createData()
+
+        # Add noise so the images are not binary
+        randState = np.random.RandomState(42)
+        doubleDonut += randState.normal(scale=0.1, size=np.shape(doubleDonut))
+        singleDonut += randState.normal(scale=0.1, size=np.shape(singleDonut))
```
teweitsai: You could put this part into the `self._createData()` as a new argument to decide to add the noise or not for the future's reusability.

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertAlmostEqual(rad, eff_radius, delta=0.1)
+
+        # Use single donut as template and test recovery on double donut
+        centX, centY, rad = self.centroidConv.getCenterAndR(
+            doubleDonut, templateDonut=singleDonut, nDonuts=2
+        )
+
+        self.assertCountEqual(centX, [50, 100])
+        self.assertEqual(centY, [80, 80])
+        eff_radius = np.sqrt(40 ** 2 - 20 ** 2)
```
teweitsai: Please add an new empty line just before `eff_radius = np.sqrt(40 ** 2 - 20 ** 2)` to improve the readability. In addition, you reused this line in several places. You could consider to put it into a new function to reduce the duplication.

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -0,0 +1,131 @@
...
+    def testGetCenterAndRFromImgBinary(self):
+
+        singleDonut, doubleDonut = self._createData()
+
+        # Test recovery with defaults
+        centX, centY, rad = self.centroidConv.getCenterAndRfromImgBinary(singleDonut)
+
+        self.assertEqual(centX, [80])
+        self.assertEqual(centY, [80])
+        eff_radius = np.sqrt(40 ** 2 - 20 ** 2)
```
teweitsai: Please add an new empty line just before `eff_radius = np.sqrt(40 ** 2 - 20 ** 2)` to improve the readability. In addition, you reused this line in several places. You could consider to put it into a new function to reduce the duplication.

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertAlmostEqual(rad, eff_radius, delta=0.1)
+
+        # Use single donut as template and test recovery on double donut
+        centX, centY, rad = self.centroidConv.getCenterAndRfromImgBinary(
+            doubleDonut, templateBinary=singleDonut, nDonuts=2
+        )
+
+        self.assertCountEqual(centX, [50, 100])
+        self.assertEqual(centY, [80, 80])
+        eff_radius = np.sqrt(40 ** 2 - 20 ** 2)
```
teweitsai: Please add an new empty line just before `eff_radius = np.sqrt(40 ** 2 - 20 ** 2)` to improve the readability. In addition, you reused this line in several places. You could consider to put it into a new function to reduce the duplication.

teweitsai (changes requested):
This update looks very good to me. Please see the comments I left. In addition, did you notice that this class will fail if someone call the function of `getImgBinary()`? You could just return the value by random walk model directly to avoid the `NotImplementedError`. Thanks!

BTW, this means you also need to add a related test case. In additional to this, in `CentroidConvolveTemplate`, you could use `self.getImgBinary(imgDonut)` instead of `self._centRandomWalk.getImgBinary(imgDonut)`. The same thing for the binary image of template. The reason is that in the future, you may want to change the method to get the binary image in `CentroidConvolveTemplate`. At that time, you could just update the function of `self.getImgBinary()`.

tests/cwfs/test_centroidConvolveTemplate.py
```diff
@@ -32,14 +32,11 @@ def setUp(self):
...
+    def _createData(self, radiusInner, radiusOuter, imageSize, addNoise=False):
 
         # Create two images. One with a single donut and one with two donuts.
-        singleDonut = np.zeros((160, 160))
-        doubleDonut = np.zeros((160, 160))
+        singleDonut = np.zeros((imageSize, imageSize))
+        doubleDonut = np.zeros((imageSize, imageSize))
 
         for x in range(160):
             for y in range(160):
```
teweitsai: Did you notice the value 160 here is `imageSize`? The similar thing such as 80 in the following. I guess 80 = `imageSize/2`?

python/lsst/ts/wep/cwfs/CentroidConvolveTemplate.py
```diff
@@ -63,35 +78,30 @@ def getCenterAndR(
...
 
-        if templateDonut is None:
-            templateDonut = deepcopy(imgDonut)
+        imgBinary = self.getImgBinary(imgDonut)
 
-        self._centRandomWalk = CentroidRandomWalk()
-        imgBinary = self._centRandomWalk.getImgBinary(imgDonut)
-        templateBinary = self._centRandomWalk.getImgBinary(templateDonut)
+        if templateDonut is None:
+            templateBinary = templateDonut
```
teweitsai: Why the `templateBinary` will be None (= `templateDonut`)?

I think you are trying to do?
```
if templateDonut is None:
    templateBinary = copy(imgBinary)
else:
    templateBinary = self.getImgBinary(templateDonut)
```
In addition, you have the following code in `getCenterAndRfromTemplateConv()` already:
```
if templateImgBinary is None:
    templateImgBinary = copy(imageBinary)
```

teweitsai (approved):
This update looks good to me. Please see the comments to fix the `getCenterAndR()` and `_createData()` before the merge. After the merge, please help to add the `tag` for the new version. Thanks for all the update!

PR #62, opened 2021-01-28
Issue/60/template generator from model

jbkalmbach:


teweitsai (changes requested):
Please base on today's discussion to update the code. Thanks!

python/lsst/ts/wep/cwfs/DonutTemplateDefault.py
```diff
@@ -0,0 +1,52 @@
...
+        defocalType : enum 'DefocalType'
+            The defocal state of the sensor.
+        imageSize : int
+            Size of template in pixels. The template will be a square.
+        **kwargs : dict[str, any]
+            Dictionary of input argument: new value for that input argument.
+
+        Returns
+        -------
+        numpy.ndarray [float]
```
teweitsai: If the return data is a binary image, why the data type is `float` instead of `int`? I thought the convention of binary image should be `int`.

python/lsst/ts/wep/cwfs/DonutTemplateFactory.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    @staticmethod
+    def createDonutTemplate(donutTemplateType):
+        """Create the template maker object.
+
+        Parameters
+        ----------
+        donutTemplateType : enum 'DonutTemplateType'
+            Algorithm to create the donut templates
```
teweitsai: Add the period in the end of description.

python/lsst/ts/wep/cwfs/DonutTemplateFactory.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Returns
+        -------
+        Child class of DonutTemplateDefault
+            Donut template object
```
teweitsai: Add the period in the end of description.

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+        Parameters
+        ----------
+        sensorName : str
+            The camera detector for which we want to make a template. Should
+            be in "Rxx_Sxx" format.
+        defocalType : enum 'DefocalType'
+            The defocal state of the sensor.
+        imageSize : int
+            Size of template in pixels. The template will be a square.
+        camType : enum 'CamType'
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+            Camera type. (Default is CamType.LsstCam)
+        model : str, optional
+            Optical model. It can be "paraxial", "onAxis", or "offAxis".
+            (The default is "offAxis")
+        pixelScale : float
```
teweitsai: This argument is "optional".

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+            The pixels to arcseconds conversion factor. (The default is 0.2)
+
+        Returns
+        -------
+        numpy.ndarray [float]
```
teweitsai: Why the return type is `float` instead of `int` for a binary image? If you call `CompensableImage.getNonPaddedMask()`, I think the output is `int`. Can you check?

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+        inst.config(camType, imageSize)
+
+        # Create image for mask
+        img = CompensableImage()
+
+        # Convert pixel locations to degrees
+        sensorXPixel = float(sensorXMicron) / pixelSizeInUm
+        sensorYPixel = float(sensorYMicron) / pixelSizeInUm
+
+        sensorXDeg = sensorXPixel * pixelScale / 3600
```
teweitsai, line 105: Maybe put a comment to explain the number of 3600 here?

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+        model : str, optional
+            Optical model. It can be "paraxial", "onAxis", or "offAxis".
+            (The default is "offAxis")
+        pixelScale : float
+            The pixels to arcseconds conversion factor. (The default is 0.2)
+
+        Returns
+        -------
+        numpy.ndarray [float]
+            The donut template as a binary image
```
teweitsai: Add the period in the end of description.

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Define position of donut at center of current sensor in degrees
+        boundaryT = 0
+        maskScalingFactorLocal = 1
+        img.setImg(fieldXY, defocalType, image=np.zeros((imageSize, imageSize)))
+        img.makeMask(inst, opticalModel, boundaryT, maskScalingFactorLocal)
+
+        templateArray = img.getNonPaddedMask()
+
+        return templateArray
```
teweitsai: You can return the value here directly because you explained it in the doc string already:
```
return img.getNonPaddedMask()
```
This is to make the code looks succinct.

tests/cwfs/test_donutTemplateModel.py
```diff
@@ -0,0 +1,83 @@
...
+        testTemplate = self._createTestTemplate(160, DefocalType.Extra, "offAxis")
+
+        # Generate a test template on the center chip
+        templateArray = self.templateMaker.makeTemplate(
+            "R22_S11", DefocalType.Extra, 160
+        )
+
+        self.assertEqual(np.max(templateArray), 1.0)
+        self.assertEqual(np.shape(templateArray), (160, 160))
+        np.testing.assert_array_equal(templateArray, testTemplate)
```
teweitsai: There is a logic problem here for the test case. It looks like the result of `_createTestTemplate()` is basically a copy of the code of `DonutTemplateModel.makeTemplate()`. Although sometimes I used the similar way to do a quick and dirty test case, I still need to ask myself that do I have any other method to generate the test case or not. Is there any physical property I could use to write the test case to a specific scientific function (not the business logic but science)? Only when I have no good idea, I will use the similar way as what you did here to generate the test case.

teweitsai (approved):

teweitsai:
The update looks good to me. Please read my comments before the merge. I have some question for the returned data type of binary image. After the merge, please add the `git tag`. Thanks!

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
jbkalmbach: The output on `CompensableImage.getNonPaddedMask()` becomes `float` after this loop: https://github.com/lsst-ts/ts_wep/blob/master/python/lsst/ts/wep/cwfs/CompensableImage.py#L1422-L1466

jbkalmbach: That is why I keep getting floats here. Should I convert my output to `int` before returning it since it is supposed to be a binary image?

tests/cwfs/test_donutTemplateModel.py
jbkalmbach: I changed the tests here. Do you mind taking another look and see if this is better?

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
jbkalmbach: I changed this to return an int since it is supposed to be binary.

teweitsai: For the link you shared, I think we did try to make the output to be `int` before:
https://github.com/lsst-ts/ts_wep/blob/master/python/lsst/ts/wep/cwfs/CompensableImage.py#L1449-L1453

Can you help to fix the bugs why the `self.cMask` and `self.pMask` are not `int` anymore? If we could fix this, we will not need to do `np.array(img.getNonPaddedMask(), dtype=int)` in `DonutTemplateModel.py`.

Thanks!

tests/cwfs/test_donutTemplateModel.py
teweitsai: I like the updated test case. Thanks!

python/lsst/ts/wep/cwfs/DonutTemplateDefault.py
teweitsai: I saw you changed this to be `int` in commit: 5aa3d68. Thanks!

python/lsst/ts/wep/cwfs/DonutTemplateModel.py
jbkalmbach: Ok, updated the code in `CompensableImage.py`

teweitsai: This looks good to me. If all Jenkins test pass, we should be good to merge the code. Thanks!

PR #64, opened 2021-02-11
Issue/61/template generator phosim

jbkalmbach:
I wasn't sure the best way to include the templates. Instead of creating all the templates and adding them to the repo I created a script to do this instead. Happy to do it another way if there's a better solution.


teweitsai (changes requested):
Please update the code based on today's discussion.

.gitignore
```diff
@@ -28,3 +28,6 @@ pytest_session.txt
 
 # Mac metadata
 .DS_Store
+
+# Existing Phosim Donut Templates
+policy/cwfs/donutTemplateData/phosimTemplates/
```
teweitsai, line 33: Please consider to add this directory by using a `.gitkeep` file.

python/lsst/ts/wep/cwfs/DonutTemplatePhosim.py
```diff
@@ -0,0 +1,92 @@
...
+            templateFilename = os.path.join(
+                phosimTemplateDir, "intra_template-%s.txt" % sensorName
+            )
+        templateArray = np.genfromtxt(templateFilename, dtype=np.int)
+
+        # Make the template the desired square shape by trimming edges of template
+        templateSize = np.shape(templateArray)
+
+        # Find the excess number of pixels in x and y direction
+        templateTrim = templateSize - np.array((imageSize, imageSize))
```
teweitsai: What will happen if `templateSize < np.array((imageSize, imageSize))`? Do you want to raise an error or do the padding instead?

python/lsst/ts/wep/cwfs/DonutTemplatePhosim.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Find the left and right edges by trimming half pixels from left and right.
+        # Then do the same for the top.
+        leftEdge = np.int(templateTrim[0] / 2)
+        rightEdge = np.int(templateSize[0] - (templateTrim[0] - leftEdge))
+        topEdge = np.int(templateTrim[1] / 2)
+        bottomEdge = np.int(templateSize[1] - (templateTrim[1] - topEdge))
+
+        templateArray = templateArray[leftEdge:rightEdge, topEdge:bottomEdge]
+        return templateArray
```
teweitsai: The updated code here is clear enough that you could do the following instead:
```
return templateArray[leftEdge:rightEdge, topEdge:bottomEdge]
```

tests/cwfs/test_donutTemplatePhosim.py
```diff
@@ -0,0 +1,118 @@
...
+        self.defaultTemplatePath = os.path.join(
+            modulePath, "policy", "cwfs", "donutTemplateData", "phosimTemplates"
+        )
+
+        # Keep track of whether directory existed before for tearDown
+        if os.path.exists(self.defaultTemplatePath):
+            self.defaultPathExists = True
+        else:
+            os.mkdir(self.defaultTemplatePath)
+            self.defaultPathExists = False
```
teweitsai: You may not need this if we create the directory in advance by `.gitkeep`. The reason to declare this first before generating the data is to let other users/developers know we might have this directory and data in advance to avoid any surprise when he/she runs the code.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import os
+import yaml
+import shutil
+import numpy as np
+from astropy.io import fits
+from lsst.ts.wep.Utility import getConfigDir, runProgram, DefocalType
+from lsst.ts.wep.CamIsrWrapper import CamIsrWrapper
+from lsst.ts.wep.cwfs.CentroidRandomWalk import CentroidRandomWalk
```
teweitsai: Please import the `CentroidFindFactory` and use the `CentroidFindFactory.createCentroidFind()` instead to get the `CentroidRandomWalk` instance. Since we are using the factory pattern now, it is easier to do the future's maintenance.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+            ts_wep/policy/cwfs/donutTemplateData/phosimTemplates.
+            (The default is "".)
+        """
+
+        self.policyDir = getConfigDir()
+        self.templateDataPath = os.path.join(
+            self.policyDir, "cwfs", "donutTemplateData"
+        )
+        self.tempWorkPath = os.path.join(self.templateDataPath, "tempDir")
+        self.repoDir = os.path.join(self.tempWorkPath, "input")
```
teweitsai: Please add the comment for each class's attributes (`self.policyDir, self.repoDir, etc.`) to explain the details. The developers are used to read the `__init__()` to know what attributes this class has. Sometimes, it might be hard to guess the meaning of a specific attribute just base on the attribute's name.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+
+        if templateDestDir == "":
+            self.templateDestDir = os.path.join(
+                self.templateDataPath, "phosimTemplates"
+            )
+        else:
+            self.templateDestDir = templateDestDir
```
teweitsai: I have the question of the design decision to let the user can assign a specific directory of templates. Does this mean we may also need to add the related parameter in `default.yaml` to tell the WEP to look for that directory in the file system? If you have the default directory: `policy/cwfs/donutTemplateData/phosimTemplates` already, why not just use it? It looks like you have put that directory into the `.gitignore` and this means the other users will not commit the local template files to the upstream by the accident. I felt this might decrease the complexity. What do you think?

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+            sensorNameFile = os.path.join(self.policyDir, "sensorNameToId.yaml")
+
+            with open(sensorNameFile, "r") as f:
+                sensorData = yaml.load(f, Loader=yaml.FullLoader)
+
+            sensorNameList = list(sensorData.keys())
+
+        else:
+
+            sensorNameList = detectorStr.split(" ")
```
teweitsai: People use the empty lines to improve the readability. But too many empty lines might hurt the readability as well.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+            String with the detector names specified in format required
+            to make flats.
+        """
+
+        if detectorStr == "":
+
+            sensorNameFile = os.path.join(self.policyDir, "sensorNameToId.yaml")
+
+            with open(sensorNameFile, "r") as f:
+                sensorData = yaml.load(f, Loader=yaml.FullLoader)
```
teweitsai: I prefer to use the `ParamReader` to get the details in the `yaml` file. But if you really want to use the `yaml` library directly, please consider to use the `yaml.safe_load()` instead of `yaml.load()`. In addition, if you just want to get all the sensor names, you could consider to use the `MapSensorNameAndId.mapSensorIdToName()` instead.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        else:
+
+            sensorNameList = detectorStr.split(" ")
+
+        detectorStrPhosim = ""
+        detectorStrFlats = ""
+
+        for sensorName in sensorNameList:
+            detectorStrPhosim += f"{sensorName}|"
+            detectorStrFlats += f"{sensorName} "
```
teweitsai: You could consider to use the `join()` to simplify the code:
https://www.geeksforgeeks.org/python-program-to-convert-a-list-to-string/

For example:
```
detectorStrPhosim = "|".join(sensorNameList)
```

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+
+        print(f"Running phosim with {numOfProc} processors")
+
+        phosimPath = os.getenv("PHOSIMPATH")
+        runPhosimCmd = f"python {phosimPath}/phosim.py"
+        runPhosimArgs = f"-w {self.tempWorkPath}/phosimWorkDir "
+        runPhosimArgs += f'-s "{detectorStrPhosim[:-1]}" '
+        runPhosimArgs += f"-p {numOfProc} "
+        runPhosimArgs += f"-i lsst "
```
teweitsai, line 142: Will we generate the "comcam" in the future? Maybe we do not need. Not sure.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+    def cutOutIntraExtraTemplates(self, templateWidth, intraVisitId, extraVisitId):
+        """
+        Cut out the donut templates from the larger extrafocal and intrafocal
+        Phosim images for every detector simulated.
+
+        Parameter
+        ---------
+        templateWidth : int
+            Width of square template image in pixels.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        intraVisitId : int
+            Visit Id of the intrafocal images from Phosim.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        Loop through all detectors in folder and cut out square region
+        around the donut from a CCD image to use
+        as the donut template for that detector. Saves the cutout to file.
+
+        Parameter
+        ---------
+        phosimImageDir : str
+            Directory where the visit ISR output is located. Inside should
+            be folders for each raft.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        templateWidth : int
+            Width of square template image in pixels.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        defocalType : enum 'DefocalType'
+            Defocal type.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        visitId : int
+            Visit Id of the defocal image from Phosim.
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        phosimTemplateDir : str, optional
+            Directory where templates will be stored. If it is an empty string
+            then it will assign it to
+            ts_wep/policy/cwfs/donutTemplateData/phosimTemplates.
+            (The default is "".)
+
```
teweitsai: This new line is not needed.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        if phosimTemplateDir == "":
+            phosimTemplateDir = os.path.join(self.templateDataPath, "phosimTemplates")
+        if phosimCentroidDir == "":
+            phosimCentroidDir = os.path.join(
+                self.tempWorkPath, "phosimOutput", defocalLabel
+            )
+
+        stampHalfWidth = int(templateWidth / 2)
+
+        centroidRW = CentroidRandomWalk()
```
teweitsai: Please use the factory method to get this instance for the future's maintenance. And use a generic variable name such as `centroidFind` instead. The benefit is to minimize the code change if you decide to use other method instead of random walk model in the future.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+
+        for sensorDir in os.listdir(phosimImageDir):
+            sensorPath = os.path.join(phosimImageDir, sensorDir)
+            for templateFile in os.listdir(sensorPath):
+                # Open ISR File
+                testHdu = fits.open(os.path.join(sensorPath, templateFile))
```
teweitsai: If you just want to get the data in fits file, you could use the `fits.getdata()` instead. You could also consider to use the `cwfs/Image.py` to get the data. I tried to restrict the use of `fits` library in that `Image` class only if people only cares about the image data instead of the header information.

tests/test_createPhosimDonutTemplates.py
```diff
@@ -0,0 +1,85 @@
...
+import unittest
+import os
+import shutil
+import numpy as np
+
+from lsst.ts.wep.CreatePhosimDonutTemplates import CreatePhosimDonutTemplates
+from lsst.ts.wep.Utility import getModulePath, DefocalType
+
+
+class TestCreatePhosimDonutTemplates(unittest.TestCase):
```
teweitsai, line 32: I think you could add other test cases such as the remove of temporary directories as well. It is good to have at least a test case for each public function if it is possible.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        Ingest flats for ISR.
+        """
+
+        print("Ingest Flats")
+
+        argString = f"{self.repoDir} "
+        argString += f"{self.tempWorkPath}/calibs/*.fits "
+        argString += f"--validity 99999 --output {self.repoDir}"
+        runProgram("ingestCalibs.py", argstring=argString)
```
teweitsai: Why you call the `ingestCalibs.py` here instead of using the `CamDataCollector`? I ask this because you used the `CamIsrWrapper` in `runISR()`.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+
+        print("Ingest images")
+
+        # Create mapper file
+        cmdString = f"echo lsst.obs.lsst.phosim.PhosimMapper > {self.tempWorkPath}/input/_mapper"
+        runProgram(cmdString)
+
+        # Run Ingestion
+        ampImg = f"{self.tempWorkPath}/raw/*.fits"
+        runProgram("ingestImages.py", argstring=f"{self.repoDir} {ampImg}")
```
teweitsai: Why you call the `ingestImages.py` here instead of using the `CamDataCollector`? I ask this because you used the `CamIsrWrapper` in `runISR()`.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def ingestImages(self):
+        """
+        Ingest the raw extrafocal images.
+        """
+
+        print("Ingest images")
+
+        # Create mapper file
+        cmdString = f"echo lsst.obs.lsst.phosim.PhosimMapper > {self.tempWorkPath}/input/_mapper"
```
teweitsai: The `CamDataCollector` has the `genPhoSimMapper()` to do this. I think the main question is that after you update the `ts_wep` to use the gen3 butler, should we keep these wrapper classes or not.

teweitsai (approved):

teweitsai:
The update looks good to me. I only have some suggestions/questions maybe you could consider.

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
teweitsai: Maybe we do not need this because comcam is similar to the central raft of lsst.

tests/test_createPhosimDonutTemplates.py
```diff
@@ -0,0 +1,200 @@
...
+        self.testDataDir = os.path.join(getModulePath(), "tests", "testData")
+        # Location where phosim input files exist.
+        # Also where temp work directory is created.
+        self.templateDataDir = os.path.join(getConfigDir(), "cwfs", "donutTemplateData")
+        # Location where templates are created
+        self.templateDir = os.path.join(self.templateDataDir, "phosimTemplates")
+        # Temporary work directory
+        self.tempWorkDir = os.path.join(self.templateDataDir, "tempDir")
+
+    def tearDown(self):
```
jbkalmbach, line 49: @teweitsai I am unsure on the best practice to design the `tearDown` when not all the methods create data that needs to be removed. 

The first `try...except` is here because only the `testCutOutTemplatesAndSave` test creates this file. 

The second `try...except` is here because I want to test the creation and clean up of the directories but if that test fails I need to be able to clean that directory still. But if it succeeds then there is nothing to clean up.

teweitsai:
I think the commit: 333f899 is a good update! Thanks for the update!

PR #78, opened 2021-04-13
Tickets/dm 29688

jbkalmbach:
Initial code to create a source catalog from pointing information using Gen3 middleware.

There are a couple notes in the code about related issues that will require us to update the code when they are merged.

This PR also includes the addition of a test repository that we can use for testing future tasks as well. It does not take up too much space and should work from the repo since I created it to use `relSymLinks` when ingesting the raws.


teweitsai:
Please do the rebase after @tribeiro merges the `tickets/DM-29687`. The docker image of `Jenkinsfile` is changed to `lsstts/develop-env`. I would like to check this PR can pass under this docker image. Thanks!

tests/testData/gen3TestRepo/convertRefCat.cfg
```diff
@@ -0,0 +1,2 @@
+config.datasetIncludePatterns = ["ref_cat", ]
+config.refCats = ["cal_ref_cat"]
```
teweitsai: Please add a new line after this. In addition, please help to add the comments to explain the details of this file and the setting used.

tests/testData/gen3TestRepo/refCat.cfg
```diff
@@ -0,0 +1,18 @@
...
+
+# Name of column to use as an identifier (optional).
+config.id_name='id'
+
+# The values in the reference catalog are assumed to be in AB magnitudes. List of column names to use for
+# photometric information.  At least one entry is required.
+config.mag_column_list=['g']
+
+# A map of magnitude column name (key) to magnitude error column (value).
+config.mag_err_column_map={'g':'g_err'}
```
teweitsai: Please add the new line in the end of file. In addition, is it possible to do something like this for the readability?
```
config.mag_err_column_map = {'g':'g_err'}
```

tests/testData/gen3TestRepo/refCat.cfg
```diff
@@ -0,0 +1,18 @@
+# String to pass to the butler to retrieve persisted files.
+# config.dataset_config.ref_dataset_name='test_refcat'
```
teweitsai: Do you need this and what is it?
```
# config.dataset_config.ref_dataset_name='test_refcat'
```

tests/testData/gen3TestRepo/refCat.cfg
```diff
@@ -0,0 +1,18 @@
...
+config.ra_name='ra'
+
+# Name of Dec column
+config.dec_name='dec'
+
+# Name of column to use as an identifier (optional).
+config.id_name='id'
+
+# The values in the reference catalog are assumed to be in AB magnitudes. List of column names to use for
+# photometric information.  At least one entry is required.
```
teweitsai, line 11:
```
# The values in the reference catalog are assumed to be in AB magnitudes.
# List of column names to use for photometric information.
# At least one entry is required.
```

In addition, what is the AB magnitude?

tests/testData/phosimOutput/realComCam/skyComCamInfoRefCatalog.txt
```diff
@@ -0,0 +1,9 @@
+id,ra,dec,g,g_err
```
teweitsai, line 1: Please add the comments to explain the details of this test file.

tests/test_generateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,126 @@
+#
```
teweitsai, line 1: Do you need this first line?
```
#
```

tribeiro:
I think you need to reformat the code with [black](https://black.readthedocs.io/en/stable/). We adopted black 20 now.

tests/test_generateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,126 @@
...
+# LSST Data Management System
```
teweitsai:
```
# This file is part of ts_wep.
#
# Developed for the LSST Telescope and Site Systems.
```

It might be better to be consistent for the license declaration. If you want to change it, please change all `python` files.

tests/testData/gen3TestRepo/butler.yaml
```diff
@@ -0,0 +1,14 @@
...
+  root: <butlerRoot>
+registry:
+  db: sqlite:///<butlerRoot>/gen3.sqlite3
+  managers:
+    attributes: lsst.daf.butler.registry.attributes.DefaultButlerAttributeManager
+    collections: lsst.daf.butler.registry.collections.synthIntKey.SynthIntKeyCollectionManager
+    datasets: lsst.daf.butler.registry.datasets.byDimensions.ByDimensionsDatasetRecordStorageManager
+    datastores: lsst.daf.butler.registry.bridge.monolithic.MonolithicDatastoreRegistryBridgeManager
+    dimensions: lsst.daf.butler.registry.dimensions.static.StaticDimensionRecordStorageManager
+    opaque: lsst.daf.butler.registry.opaque.ByNameOpaqueTableStorageManager
```
teweitsai, line 14: Please add the comments to explain each items in this `yaml` file.

tests/testData/createGen3TestRepo.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Make Gen 2 Repo for ref cat
+mkdir gen2TestRepo
+echo "lsst.obs.lsst.LsstCamMapper" > gen2TestRepo/_mapper
+ingestReferenceCatalog.py gen2TestRepo phosimOutput/realComCam/skyComCamInfoRefCatalog.txt --configfile gen3TestRepo/refCat.cfg
+# Convert to Gen 3 (currently only way to get Gen3 ref cat)
```
teweitsai, line 8: Please add a new/empty line before this for the readability.

tests/testData/createGen3TestRepo.sh
```diff
@@ -0,0 +1,15 @@
...
+butler convert --gen2root gen2TestRepo --config-file gen3TestRepo/convertRefCat.cfg gen3TestRepo
+# Ingest Raws
```
teweitsai, line 11: Please add a new/empty line before this for the readability.

tests/testData/createGen3TestRepo.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Make Gen 2 Repo for ref cat
```
teweitsai: reference catalog?

tests/testData/createGen3TestRepo.sh
```diff
@@ -0,0 +1,15 @@
...
+ingestReferenceCatalog.py gen2TestRepo phosimOutput/realComCam/skyComCamInfoRefCatalog.txt --configfile gen3TestRepo/refCat.cfg
+# Convert to Gen 3 (currently only way to get Gen3 ref cat)
+butler convert --gen2root gen2TestRepo --config-file gen3TestRepo/convertRefCat.cfg gen3TestRepo
+# Ingest Raws
+butler ingest-raws -t relsymlink gen3TestRepo phosimOutput/realComCam/repackagedFiles/extra/*.fits
+butler ingest-raws -t relsymlink gen3TestRepo phosimOutput/realComCam/repackagedFiles/intra/*.fits
+butler define-visits gen3TestRepo lsst.obs.lsst.LsstCam
+
+# Clean Up
+rm -r gen2TestRepo
```
teweitsai: Please add a new line after this.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
+#
+# LSST Data Management System
+# Copyright 2008-2017 AURA/LSST.
```
teweitsai:
```
# This file is part of ts_wep.
#
# Developed for the LSST Telescope and Site Systems.
```

It might be better to be consistent for the license declaration. If you want to change it, please change all `python` files.

tests/testData/gen3TestRepo/butler.yaml
jbkalmbach: This is the file created automatically by the butler when making the repository. I didn't create these settings so I'm not sure what they all mean.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+import lsst.afw.table as afwTable
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as connectionTypes
+import lsst.obs.lsst as obs_lsst
+from lsst.meas.algorithms import ReferenceObjectLoader, LoadReferenceObjectsConfig
+from lsst.obs.base import createInitialSkyWcsFromBoresight
+
+
+class GenerateDonutCatalogOnlineTaskConnections(
```
teweitsai: Please add the doc string for this class. In addition, please update the `doc/wepClass.uml` and `doc/content.rst` as well. BTW, I prefer each class has its own python file. This is easier to maintain and know which python file to update the code. Or you could have a new module in `ts_wep` to put the pipeline tasks. This might be easier to maintain.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+    )
+    donutCatalog = connectionTypes.Output(
+        doc="Donut Locations",
+        dimensions=("instrument",),
+        storageClass="DataFrame",
+        name="donutCatalog",
+    )
+
+
+class GenerateDonutCatalogOnlineTaskConfig(
```
teweitsai: Please add the doc string for this class. In addition, please update the `doc/wepClass.uml` and `doc/content.rst` as well. BTW, I prefer each class has its own python file. This is easier to maintain and know which python file to update the code. Or you could have a new module in `ts_wep` to put the pipeline tasks. This might be easier to maintain.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+    boresightDec = pexConfig.Field(
+        doc="Boresight Dec in degrees", dtype=float, default=0.0
+    )
+    boresightRotAng = pexConfig.Field(
+        doc="Boresight Rotation Angle in degrees", dtype=float, default=0.0
+    )
+    cameraName = pexConfig.Field(doc="Camera Name", dtype=str, default="lsstCam")
+
+
+class GenerateDonutCatalogOnlineTask(pipeBase.PipelineTask):
```
teweitsai: Please update the `doc/wepClass.uml` and `doc/content.rst` as well. BTW, I prefer each class has its own python file. This is easier to maintain and know which python file to update the code. Or you could have a new module in `ts_wep` to put the pipeline tasks. This might be easier to maintain.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+    ConfigClass = GenerateDonutCatalogOnlineTaskConfig
+    _DefaultName = "GenerateDonutCatalogOnlineTask"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.filterName = self.config.filterName
+        self.boresightRa = self.config.boresightRa
+        self.boresightDec = self.config.boresightDec
+        self.boresightRotAng = self.config.boresightRotAng
+        self.cameraName = self.config.cameraName
```
teweitsai: Please add the comments to explain each attributes, especially for the attributes that are physical parameters contain the units and coordinate system.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # TODO: Temporary until DM-24162 is closed at which point we
+        # can remove this
+        os.environ["NUMEXPR_MAX_THREADS"] = "1"
```
teweitsai: Please consider to put this into the configuration file as such the `policy/default.yaml`. You could see the `numOfProc` is there already. If the pipeline task is based on the `thread`, maybe rename the `numOfProc` to `numOfThread` and modify the related description.

PS. In Linux, there should be no much difference between thread and processor in the internal realization.

tests/testData/gen3TestRepo/refCat.cfg
jbkalmbach: AB magnitude is just the astronomical magnitude system expected by the LSST DM code.

https://en.wikipedia.org/wiki/AB_magnitude

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def filterResults(self, resultsDataFrame):
+        """
+        Here is where we will set up specifications for the sources
+        we want to use (i.e., filter on magnitude, blended, etc.).
+        For now it just returns all sources.
+        """
```
teweitsai: Please provide the doc string of input/ouput argument: `resultsDataFrame`. BTW, I felt the doc string you put is more like the comments of this function instead of its purpose. Please consider to rewrite it.

tests/testData/phosimOutput/realComCam/skyComCamInfoRefCatalog.txt
jbkalmbach: Is there someplace else I can put the explanation for this file? If I add the comments here the file does not get ingested by the Butler properly.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+
+        return resultsDataFrame
+
+    def run(
+        self,
+        refCatalogs: typing.List[afwTable.SimpleCatalog],
+    ) -> pipeBase.Struct:
```
teweitsai: I think this function is inherit from the parent and you do not put the doc string. Is this correct?

jbkalmbach:
> I think you need to reformat the code with [black](https://black.readthedocs.io/en/stable/). We adopted black 20 now.

I did using version `20.8b1`. What files look like they were missed?

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+        # Set up pandas dataframe
+        fieldObjects = pd.DataFrame([])
+        ra = []
+        dec = []
+        centroidX = []
+        centroidY = []
+        det_names = []
+
+        # Get camera. Only 'lsstCam' for now.
+        if self.cameraName == "lsstCam":
```
teweitsai: I think it is better to use the `Utility.CamType` and the user could put this in the runtime or read from the configuration file (`camMapper` in `policy/default.yaml`). You could raise the error if that camera is not supported yet. It is better to not hard-coded the camera here.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
jbkalmbach: Yes, that is correct.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+            lsst_cam = obs_lsst.LsstCam.getCamera()
+        else:
+            raise ValueError(f"{self.cameraName} is not a valid camera name.")
+
+        for detector in lsst_cam:
```
teweitsai: We do not want to hard-coded the camera here (`lsst_cam`). You could do the following as an example:
```
camera = obs_lsst.LsstCam.getCamera()
for detector in camera:
```

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+            # Create WCS from boresight information
+            # NOTE: 90.0 - boresightRotAng is a phosim artifact that is being fixed
+            # TODO: Fix this when DM-29702 is merged
+            detWcs = createInitialSkyWcsFromBoresight(
+                lsst.geom.SpherePoint(
+                    self.boresightRa, self.boresightDec, lsst.geom.degrees
+                ),
+                (90.0 - self.boresightRotAng) * lsst.geom.degrees,
+                detector,
+            )
```
teweitsai: I felt the code here is a duplication of `WcsSol` (you have `90.0` in both side). I prefer you use the `WcsSol` instead. Then, when you need to fix this, you only need to fix one part of code. The idea is to reduce the duplication as more as possible.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,169 @@
...
+        # Flatten information from all detector lists and enter into dataframe
+        fieldObjects["coord_ra"] = np.hstack(ra).squeeze()
+        fieldObjects["coord_dec"] = np.hstack(dec).squeeze()
+        fieldObjects["centroid_x"] = np.hstack(centroidX).squeeze()
+        fieldObjects["centroid_y"] = np.hstack(centroidY).squeeze()
+        fieldObjects["detector"] = np.hstack(det_names).squeeze()
+
+        finalSources = self.filterResults(fieldObjects)
+
+        return pipeBase.Struct(donutCatalog=finalSources)
```
teweitsai: It might be good to explain the output if you do not have the related doc string.

teweitsai (changes requested):
1. Please update the `doc/versionHistory.rst`.
2. Is it possible to create the following file in the test instead of putting into the repo?
`tests/testData/gen3TestRepo/gen3.sqlite3`
3. Please fix the warnings in Jenkins test.

tests/testData/gen3TestRepo/refCat.cfg
teweitsai: Is it normal to write the abbreviation? I prefer the full name actually because the developers/maintainers may not be in this field (at least, I am not).

tests/testData/phosimOutput/realComCam/skyComCamInfoRefCatalog.txt
teweitsai: Do you mean you could not do something like this?
```
# this is the explanation
id,ra,dec,g,g_err
```

If this is the case, maybe explain it a little bit in `README.md`.

tests/testData/gen3TestRepo/butler.yaml
teweitsai: Can you help to explain this part in `README.md`?

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
jbkalmbach: This is only temporary until that issue gets closed. This number should always be 1 inside this pipe task or there will be problems with the multiprocessing in the quantum graph structure.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
jbkalmbach: I would rather have it come from the `GenerateDonutCatalogOnlineTaskConfig` class so we can keep everything for this Task in one place. I think this will be especially useful when running as a pipeline task since the configurations can all be specified in the pipeline `yaml` file used to specify the tasks and configurations in a pipeline.

tests/testData/gen3TestRepo/refCat.cfg
jbkalmbach: It's always just called the AB magnitude system. I think it's not even an abbreviation but historically short for "absolute" something...calibration maybe...but I don't actually know.

python/lsst/ts/wep/GenerateDonutCatalogOnlineTask.py
teweitsai: Yes. I think you are right. This comes from `self.cameraName = self.config.cameraName`. Thanks!

jbkalmbach:
> 1. Please update the `doc/versionHistory.rst`.
> 2. Is it possible to create the following file in the test instead of putting into the repo?
>    `tests/testData/gen3TestRepo/gen3.sqlite3`
> 3. Please fix the warnings in Jenkins test.

I added information for `gen3.sqlite3` in the documentation but it does need to be there for the Gen 3 Test Repo to work.

I don't think any of the warning in the Jenkins test are from the code I added. I think the `np.bool` warnings are because the version of `skimage` in the stack needs to be updated. Perhaps we need a new issue+PR to take care of the other ones?

README.md
```diff
@@ -123,6 +123,21 @@ setup -k -r .
 1. While testing with the sky images obtained with the auxiliary telescope (localed in `tests/testData/testImages/auxTel`), this package and [cwfs](https://github.com/bxin/cwfs) show the similar result in "onAxis" optical model. However, for the "paraxial" optical model, the results of two packages are different.
 2. The main difference comes from the stratege of compensation in the initial loop of solving the TIE. However, it is hard to have a conclusion at this moment because of the low singal-to-noise ratio in test images.
 
+## Test Gen 3 Repository
+
+In the folder `tests/testData/` there is a test repository for tasks that run with the Gen 3 DM middleware. This repository is the folder `tests/testData/gen3TestRepo` and was built using the
+files `tests/testData/createGen3TestRepo.sh`. This script performs the following steps to create the Gen 3 repository:
```
teweitsai: Please put each sentence as a new line for the benefit of `git diff` in future:

In the folder `tests/testData/` there is a test repository for tasks that run with the Gen 3 DM middleware.
This repository is the folder `tests/testData/gen3TestRepo` and was built using the files `tests/testData/createGen3TestRepo.sh`.
This script performs the following steps to create the Gen 3 repository:

README.md
```diff
@@ -123,6 +123,21 @@ setup -k -r .
...
+
+  1. Ingest a reference catalog with the Gen 2 middleware. This reference catalog is the file `tests/testData/phosimOutput/realComCam/skyComCamInfoRefCatalog.txt` which contains the same information as
+  `skyComCamInfo.txt` in the same folder but formatted to be read in by the Gen 2 middleware as a reference catalog. A column of g magnitude error is added with a value of 0.1 to fit reference catalog format.
+  The format of the reference catalog is configured with the file `tests/testData/gen3TestRepo/refCat.cfg`.
+  2. Convert the Gen 2 repository with the reference catalog into a Gen 3 repository since this is currently the only way to create a Gen 3 reference catalog. This requires the configuration file
+  `tests/testData/gen3TestRepo/convertRefCat.cfg`.
+  3. Ingest the raw files in `tests/testData/phosimOutput/realComCam/repackagedFiles`.
+  4. Clean up the original Gen 2 repository.
```
teweitsai: Please put each sentence as a new line for the benefit of `git diff` in future. You could do some experiments and check the format will keep the same even the new line is used. In addition, you may not need the space before each point (1, 2, 3, 4).

python/lsst/ts/wep/task/GenerateDonutCatalogOnlineTask.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class GenerateDonutCatalogOnlineTask(pipeBase.PipelineTask):
+    """
+    Create a WCS from boresight info and then use this
+    with a reference catalog to select sources on the detectors for AOS.
+    """
+
+    ConfigClass = GenerateDonutCatalogOnlineTaskConfig
+    _DefaultName = "GenerateDonutCatalogOnlineTask"
```
teweitsai: Does the `_DefaultName` need to be the same as class name? If this is the case, should we comment this or find a way to get this attribute's value programmatically?

doc/uml/taskClass.uml
```diff
@@ -0,0 +1,4 @@
+@startuml
+GenerateDonutCatalogOnlineTaskConnections --* GenerateDonutCatalogOnlineTaskConfig
+GenerateDonutCatalogOnlineTaskConfig --* GenerateDonutCatalogOnlineTask
```
teweitsai: This should be:
```
GenerateDonutCatalogOnlineTaskConfig *-- GenerateDonutCatalogOnlineTaskConnections 
GenerateDonutCatalogOnlineTask *-- GenerateDonutCatalogOnlineTaskConfig
```
Do you see the difference in the generated figure?

doc/uml/taskClass.uml
```diff
@@ -0,0 +1,4 @@
...
+@enduml
```
teweitsai: Please add a new line after this.

teweitsai (approved):
This PR looks good to me. I have some comments that you could consider before the merge. Thanks!

python/lsst/ts/wep/task/GenerateDonutCatalogOnlineTask.py
jbkalmbach: It doesn't need to be the same as the class name but I think you need to give it a descriptive name for the pipeline maybe. All the other pipeline tasks I've been using as a template have something similar to the class name here.

PR #79, opened 2021-04-13
Tickets/dm 29687

tribeiro:


teweitsai (approved):
This PR looks good to me. However, we may need to consider that the `ts_wep` update of gen3 might need the latest `obs_lsst` to support. The update frequency of `lsstts/develop-env` is much slower than `lsstsqre/centos:w_latest`.

tribeiro:
uhm.. actually `lsstts/develop-env:develop` updates synchronously with `lsstsqre/centos:w_latest`, assuming nothing breaks... but this is a good point... it may actually be a good think to transition to `lsstts/develop-env:develop` since this is our true baseline.

PR #80, opened 2021-04-19
[DM-29787] Get instrument from pipeline in DonutCatalogOnlineTask

jbkalmbach:
In the new `GenerateDonutCatalogOnlineTask` in ts_wep the camera to set up the WCS is a configuration parameter. But I recently learned how to make it so that it will come from the pipeline settings. This is a better way to do things so it removes the possibility of having the camera in the pipeline be different from the camera set up to create the donutCatalog with X,Y pixel locations.

### DIFF
--- a/tests/task/test_generateDonutCatalogOnlineTask.py
+++ b/tests/task/test_generateDonutCatalogOnlineTask.py
@@ -45,6 +45,18 @@ class TestGenerateDonutCatalogOnlineTask(unittest.TestCase):
         self.butler = dafButler.Butler(self.repoDir)
         self.registry = self.butler.registry
 
+    def _writePipetaskCmd(self, runName, taskName):
+
+        pipetaskCmd = "pipetask run "
+        pipetaskCmd += f"-b {self.repoDir} "  # Specify repo
+        pipetaskCmd += "-i refcats "  # Specify collections with data to use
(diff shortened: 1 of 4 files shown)


tests/task/test_generateDonutCatalogOnlineTask.py
```diff
@@ -68,6 +67,65 @@ def testFilterResults(self):
...
+        pipetaskCmd = "pipetask run "
+        pipetaskCmd += f"-b {self.repoDir} "  # Specify repo
+        pipetaskCmd += "-i refcats "  # Specify collections with data to use
+        # Specify task
+        taskName = "GenerateDonutCatalogOnlineTask.GenerateDonutCatalogOnlineTask"
+        pipetaskCmd += f"-t lsst.ts.wep.task.{taskName} "
+        pipetaskCmd += "--instrument lsst.obs.lsst.LsstCam "
+        pipetaskCmd += f"--register-dataset-types --output-run {runName}"
+
+        runProgram(pipetaskCmd)
```
teweitsai: Please consider to abstract this part as a new function for the readability.

teweitsai (approved):
The update looks good to me. Please help to fix the failed Jenkins test. Thanks!

PR #81, opened 2021-04-21
DM-29811: Update ROTANG header in test files

suberlak:
Updated header in test files to match the WCS information, following  an update to phosim_utils via https://jira.lsstcorp.org/browse/DM-29702


teweitsai (approved):
Please help to update the `doc/versionHistory.rst` before the merge. Thanks!

In addition, I am considering the following code will be affected or not:
https://github.com/lsst-ts/ts_wep/blob/develop/python/lsst/ts/wep/bsc/WcsSol.py#L94-L99

You could see there is a `90.0` degree offset there.

PR #83, opened 2021-04-27
Tickets/dm 29789

jbkalmbach:
This will take the donutCatalog from GenerateDonutCatalogOnlineTask and the post-ISR exposures to run the Zernike estimation. It will save donut stamps and Zernike coefficients in the repository for access through the Gen 3 butler.

### DIFF
--- a/doc/uml/taskClass.uml
+++ b/doc/uml/taskClass.uml
@@ -2,4 +2,8 @@
 GenerateDonutCatalogOnlineTaskConfig *-- GenerateDonutCatalogOnlineTaskConnections
 GenerateDonutCatalogOnlineTask *-- GenerateDonutCatalogOnlineTaskConfig
 DonutStamps *-- DonutStamp
+EstimateZernikesTaskConfig *-- EstimateZernikesTaskConnections
+EstimateZernikesTask *-- EstimateZernikesTaskConfig
+EstimateZernikesTask *-- DonutStamps
+EstimateZernikesTask *-- DonutStamp
 @enduml
--- a/doc/uml/taskClass.uml
+++ b/doc/uml/taskClass.uml
@@ -2,8 +2,8 @@
 GenerateDonutCatalogOnlineTaskConfig *-- GenerateDonutCatalogOnlineTaskConnections
 GenerateDonutCatalogOnlineTask *-- GenerateDonutCatalogOnlineTaskConfig
 DonutStamps *-- DonutStamp
-EstimateZernikesTaskConfig *-- EstimateZernikesTaskConnections
-EstimateZernikesTask *-- EstimateZernikesTaskConfig
-EstimateZernikesTask *-- DonutStamps
-EstimateZernikesTask *-- DonutStamp
+EstimateZernikesFamTaskConfig *-- EstimateZernikesFamTaskConnections
+EstimateZernikesFamTask *-- EstimateZernikesFamTaskConfig
+EstimateZernikesFamTask ..> DonutStamps
+EstimateZernikesFamTask ..> DonutStamp
 @enduml
--- a/python/lsst/ts/wep/Utility.py
+++ b/python/lsst/ts/wep/Utility.py
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
         raw amplifier image files
     """
     return list(filter(re.compile(r"^((?!_e).)*fits$").match, os.listdir(rawExpDir)))
+
+
+def writePipetaskCmd(
+    repoDir, runName, instrument, collections, taskName=None, pipelineYaml=None
+):
+    """
+    Format a command line call to run a Gen 3 pipeline task.
+    Can also be a set of tasks if specified in a pipeline yaml file.
+
+    Parameters
+    ----------
+    repoDir: str
+        Location of Gen 3 repository
+    runName: str
+        Name of collection for data produced by the task
+    instrument: str
+        The instrument to use for the task.
+    collections: str
+        The data collections needed for the task.
+    taskName: str, optional
+        Full task function name in lsst namespace. One of taskName
+        or pipelineYaml must be specified to run. (The default is None).
+    pipelineYaml: str, optional
+        Yaml file that specifies a pipeline configuration to run
+        instead of a single task. (The default is None.)
+
+    Returns
+    -------
+    str
+        Pipetask run command
+    """
+    assertMsg = "At least one of taskName or pipelineYaml must not be None"
+    if (taskName is None) & (pipelineYaml is None):
+        raise ValueError(assertMsg)
+
+    pipetaskCmd = "pipetask run "
+    pipetaskCmd += f"-b {repoDir} "  # Specify repo
+    pipetaskCmd += f"-i {collections} "  # Specify collections with data to use
+    pipetaskCmd += f"--instrument {instrument} "
+    pipetaskCmd += f"--register-dataset-types --output-run {runName}"
+    if taskName is not None:
+        pipetaskCmd += f" -t {taskName}"
+    if pipelineYaml is not None:
+        pipetaskCmd += f" -p {pipelineYaml}"
+
+    return pipetaskCmd
+
+
+def writeCleanUpRepoCmd(repoDir, runName):
+    """
+    Format a command line call to clean up the data created by a pipeline.
+
+    Parameters
+    ----------
+    repoDir: str
+        Location of Gen 3 repository
+    runName: str
+        Name of collection for data produced by the task
+
+    Returns
+    -------
+    str
+        Butler prune-collection command
+    """
+
+    cleanUpCmd = "butler prune-collection "
+    cleanUpCmd += f"{repoDir} {runName} --purge --unstore"
+
+    return cleanUpCmd
--- a/python/lsst/ts/wep/Utility.py
+++ b/python/lsst/ts/wep/Utility.py
@@ -470,9 +470,9 @@ def writePipetaskCmd(
     Parameters
     ----------
     repoDir: str
-        Location of Gen 3 repository
+        Location of Gen 3 repository.
     runName: str
-        Name of collection for data produced by the task
+        Name of collection for data produced by the task.
     instrument: str
         The instrument to use for the task.
     collections: str
@@ -487,11 +487,15 @@ def writePipetaskCmd(
     Returns
     -------
     str
-        Pipetask run command
+        Pipetask run command.
+
+    Raises
+    ------
+    ValueError
+        Need to at least specify name of task or name of pipeline file.
     """
-    assertMsg = "At least one of taskName or pipelineYaml must not be None"
-    if (taskName is None) & (pipelineYaml is None):
-        raise ValueError(assertMsg)
+    if (taskName is None) and (pipelineYaml is None):
+        raise ValueError("At least one of taskName or pipelineYaml must not be None")
 
     pipetaskCmd = "pipetask run "
     pipetaskCmd += f"-b {repoDir} "  # Specify repo
@@ -513,14 +517,14 @@ def writeCleanUpRepoCmd(repoDir, runName):
     Parameters
     ----------
     repoDir: str
-        Location of Gen 3 repository
+        Location of Gen 3 repository.
     runName: str
-        Name of collection for data produced by the task
+        Name of collection for data produced by the task.
 
     Returns
     -------
     str
-        Butler prune-collection command
+        Butler prune-collection command.
     """
 
     cleanUpCmd = "butler prune-collection "
--- a/python/lsst/ts/wep/task/EstimateZernikesFamTask.py
+++ b/python/lsst/ts/wep/task/EstimateZernikesFamTask.py
@@ -23,7 +23,6 @@ import os
 import typing
 import numpy as np
 import pandas as pd
-from copy import copy
 
 import lsst.geom
 import lsst.pipe.base as pipeBase
@@ -150,6 +149,7 @@ class EstimateZernikesFamTask(pipeBase.PipelineTask):
 
     def shiftCenter(self, center, boundary, distance):
         """Shift the center if its distance to boundary is less than required.
+
         Parameters
         ----------
         center : float
@@ -158,6 +158,7 @@ class EstimateZernikesFamTask(pipeBase.PipelineTask):
             Boundary point.
         distance : float
             Required distance.
+
         Returns
         -------
         float
@@ -241,11 +242,9 @@ class EstimateZernikesFamTask(pipeBase.PipelineTask):
         finalDonutY = yCent + (newY - initialHalfWidth)
 
         # Shift stamp center if necessary but not final centroid definition
-        xStampCent = copy(finalDonutX)
-        yStampCent = copy(finalDonutY)
-        xStampCent = self.shiftCenter(xStampCent, expDim.getX(), stampHalfWidth)
+        xStampCent = self.shiftCenter(finalDonutX, expDim.getX(), stampHalfWidth)
         xStampCent = self.shiftCenter(xStampCent, 0, stampHalfWidth)
-        yStampCent = self.shiftCenter(yStampCent, expDim.getY(), stampHalfWidth)
+        yStampCent = self.shiftCenter(finalDonutY, expDim.getY(), stampHalfWidth)
         yStampCent = self.shiftCenter(yStampCent, 0, stampHalfWidth)
 
         # Define corner for final stamp BBox
@@ -296,6 +295,10 @@ class EstimateZernikesFamTask(pipeBase.PipelineTask):
             xCent = int(donutRow["centroid_x"])
             yCent = int(donutRow["centroid_y"])
 
+            # Adjust the centroid coordinates from the catalog by convolving
+            # the postage stamp with the donut template and return
+            # the new centroid position as well as the corners of the
+            # postage stamp to cut out of the exposure.
             finalDonutX, finalDonutY, xCorner, yCorner = self.calculateFinalCentroid(
                 exposure, template, xCent, yCent
             )
--- a/python/lsst/ts/wep/task/EstimateZernikesFamTask.py
+++ b/python/lsst/ts/wep/task/EstimateZernikesFamTask.py
@@ -23,6 +23,7 @@ import os
 import typing
 import numpy as np
 import pandas as pd
+from copy import copy
 
 import lsst.geom
 import lsst.pipe.base as pipeBase
@@ -41,7 +42,7 @@ from lsst.ts.wep.task.DonutStamps import DonutStamp, DonutStamps
 from lsst.pipe.base import connectionTypes
 
 
-class EstimateZernikesTaskConnections(
+class EstimateZernikesFamTaskConnections(
     pipeBase.PipelineTaskConnections, dimensions=("detector", "instrument")
 ):
     exposures = connectionTypes.Input(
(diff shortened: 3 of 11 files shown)


python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+    repoDir, runName, instrument, collections, taskName=None, pipelineYaml=None
+):
+    """
+    Format a command line call to run a Gen 3 pipeline task.
+    Can also be a set of tasks if specified in a pipeline yaml file.
+
+    Parameters
+    ----------
+    repoDir: str
+        Location of Gen 3 repository
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+    runName: str
+        Name of collection for data produced by the task
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+        instead of a single task. (The default is None.)
+
+    Returns
+    -------
+    str
+        Pipetask run command
+    """
+    assertMsg = "At least one of taskName or pipelineYaml must not be None"
+    if (taskName is None) & (pipelineYaml is None):
+        raise ValueError(assertMsg)
```
teweitsai: Please add the doc string of raised error. In addition, why not?
```
raise ValueError("At least one of taskName or pipelineYaml must not be None")
```

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+        Yaml file that specifies a pipeline configuration to run
+        instead of a single task. (The default is None.)
+
+    Returns
+    -------
+    str
+        Pipetask run command
+    """
+    assertMsg = "At least one of taskName or pipelineYaml must not be None"
+    if (taskName is None) & (pipelineYaml is None):
```
teweitsai: Why not?
```
if (taskName is None) and (pipelineYaml is None)
```

`&` is bit-and in the most of languages:
https://stackoverflow.com/questions/2485466/pythons-equivalent-of-logical-and-in-an-if-statement

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+
+
+def writeCleanUpRepoCmd(repoDir, runName):
+    """
+    Format a command line call to clean up the data created by a pipeline.
+
+    Parameters
+    ----------
+    repoDir: str
+        Location of Gen 3 repository
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+    runName: str
+        Name of collection for data produced by the task
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/Utility.py
```diff
@@ -458,3 +458,72 @@ def getAmpImagesFromDir(rawExpDir):
...
+
+    Returns
+    -------
+    str
+        Butler prune-collection command
```
teweitsai: Period in the end of sentence.

tests/test_utility.py
```diff
@@ -149,6 +151,46 @@ def testGetAmpImagesFromDir(self):
...
+
+        testCmd = f"pipetask run -b {repoName} -i {collections} "
+        testCmd += f"--instrument {instrument} "
+        testCmd += f"--register-dataset-types --output-run {runName}"
+        testCmdTask = testCmd + f" -t {taskName}"
+        testCmdYaml = testCmd + f" -p {pipelineYamlFile}"
+
+        pipeOutTask = writePipetaskCmd(
+            repoName, runName, instrument, collections, taskName=taskName
+        )
```
teweitsai: Please consider to abstract this part as a function to improve the readability. For example, `self._writePipetaskCmd()` or something similar.

tests/test_utility.py
```diff
@@ -149,6 +151,46 @@ def testGetAmpImagesFromDir(self):
...
+            writePipetaskCmd(repoName, runName, instrument, collections)
+        self.assertTrue(assertMsg in str(context.exception))
+
+    def testWriteCleanUpRepoCmd(self):
+
+        repoName = "testRepo"
+        runName = "run2"
+
+        testCmd = f"butler prune-collection {repoName} {runName}"
+        testCmd += " --purge --unstore"
```
teweitsai: Please consider to abstract this part as a function to improve the readability.

tests/testData/testTaskPipeline.yaml
```diff
@@ -0,0 +1,28 @@
+description: wep basic processing pipeline
```
teweitsai: Please consider to write some comments to explain this test yaml file. I am not familiar with the gen 3 butler and need to learn from you. Thanks!

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+
+class EstimateZernikesTaskConnections(
+    pipeBase.PipelineTaskConnections, dimensions=("detector", "instrument")
+):
+    exposures = connectionTypes.Input(
+        doc="Input exposure to make measurements on",
+        dimensions=("exposure", "detector", "instrument"),
+        storageClass="Exposure",
+        name="postISRCCD",
+        multiple=True,
```
teweitsai: What is the meaning of `multiple` here?

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+class EstimateZernikesTaskConfig(
+    pipeBase.PipelineTaskConfig, pipelineConnections=EstimateZernikesTaskConnections
+):
+    donutTemplateSize = pexConfig.Field(doc="Size of Template", dtype=int, default=160)
+    donutStampSize = pexConfig.Field(doc="Size of donut stamps", dtype=int, default=160)
+    initialCutoutSize = pexConfig.Field(
+        doc="Size of initial donut cutout used to centroid", dtype=int, default=240
+    )
```
teweitsai: How do I change the default value (160 or 240)? By the configuration yaml file? Or by the `Config` class in DM team?

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+    ConfigClass = EstimateZernikesTaskConfig
+    _DefaultName = "EstimateZernikesTask"
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        # Set up configs
+        self.donutTemplateSize = self.config.donutTemplateSize
+        self.donutStampSize = self.config.donutStampSize
+        self.initialCutoutSize = self.config.initialCutoutSize
```
teweitsai: Please write some comments to explain these three attributes for other developers to understand. For me, I am not sure I understand the meaning of `self.initialCutoutSize`.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+    def getTemplate(self, detectorName, defocalType):
+        """
+        Get the templates for the detector
+
+        Parameters
+        ----------
+        detectorName: str
+            Name of the CCD (e.g. 'R22_S11').
+        defocalType: enum 'DefocalType'
+            Defocal type of the donut image
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+        super().__init__(**kwargs)
+
+        # Set up configs
+        self.donutTemplateSize = self.config.donutTemplateSize
+        self.donutStampSize = self.config.donutStampSize
+        self.initialCutoutSize = self.config.initialCutoutSize
+
+    def getTemplate(self, detectorName, defocalType):
+        """
+        Get the templates for the detector
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+        Cut out postage stamps for sources in catalog.
+
+        Parameters
+        ----------
+        exposure: afwImage.Exposure
+            Post-ISR image with defocal donuts sources.
+        donutCatalog: pandas DataFrame
+            Source catalog for the pointing.
+        defocalType: enum 'DefocalType'
+            Defocal type of the donut image
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+
+        return template
+
+    def cutOutStamps(self, exposure, donutCatalog, defocalType):
+        """
+        Cut out postage stamps for sources in catalog.
+
+        Parameters
+        ----------
+        exposure: afwImage.Exposure
```
teweitsai: `lsst.afw.image.Exposure`? It might be better to write down the full path (the user may not know `import lsst.afw.image as afwImage`).

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+            Post-ISR image with defocal donuts sources.
+        donutCatalog: pandas DataFrame
+            Source catalog for the pointing.
+        defocalType: enum 'DefocalType'
+            Defocal type of the donut image
+
+        Returns
+        -------
+        DonutStamps
+            Collection of postage stamps as afwImage.MaskedImages
```
teweitsai: `lsst.afw.image.MaskedImages`? It might be better to write down the full path (the user may not know `import lsst.afw.image as afwImage`).

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+        template = self.getTemplate(detectorName, defocalType)
+        detectorCatalog = donutCatalog.query(f'detector == "{detectorName}"')
+
+        initialHalfWidth = int(self.initialCutoutSize / 2)
+        stampHalfWidth = int(self.donutStampSize / 2)
+        finalStamps = []
+        finalXList = []
+        finalYList = []
+        xLowList = []
+        yLowList = []
```
teweitsai: What are the `xLowList` and `yLowList`? This means please consider to add some comments to explain them.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+
+        for donutRow in detectorCatalog.to_records():
+            # Make an initial cutout larger than the actual final stamp
+            # so that we can centroid to get the stamp centered exactly
+            # on the donut
+            xCent = int(donutRow["centroid_x"])
+            yCent = int(donutRow["centroid_y"])
+            initCutoutHalfWidth = self.initialCutoutSize / 2
+            # BBox measured from corner of stamp not center
```
teweitsai: I felt this sentence is a little weird. BBox measured from corner of stamp that is not center?

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+            # Get the final cutout
+            xLow = finalDonutX - stampHalfWidth
+            yLow = finalDonutY - stampHalfWidth
+            finalCent = lsst.geom.Point2I(xLow, yLow)
+            finalBBox = lsst.geom.Box2I(
+                finalCent, lsst.geom.Extent2I(self.donutStampSize)
+            )
+            xLowList.append(xLow)
+            yLowList.append(yLow)
+            finalCutout = exposure[finalBBox]
```
teweitsai: Do you consider the condition that calculated x, y dimension of cutout image may be more than the CCD (4000 x 4000) itself? For example, the center of maximum correlation (x, y) = (1000, 3950) and the stamp size you want to cut is 200 x 200. At this time, 3950 + 200/2 = 4050 > 4000.

This is what I did before:
https://github.com/lsst-ts/ts_wep/blob/develop/python/lsst/ts/wep/SourceProcessor.py#L659-L665

The idea is to shift the center of stamp to be able to cut the image.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+        return DonutStamps(finalStamps, metadata=stampsMetadata)
+
+    def estimateZernikes(self, donutStampsExtra, donutStampsIntra):
+        """
+        Take the donut postage stamps and estimate the Zernike coefficients.
+
+        Parameters
+        ----------
+        donutStampsExtra: DonutStamps
+            Extra-focal donut postage stamps
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+        donutStampsIntra: DonutStamps
+            Intra-focal donut postage stamps
```
teweitsai: Period in the end of sentence.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+
+        Returns
+        -------
+        numpy.ndarray
+            Zernike coefficients for the exposure. Will return one set of
+            coefficients per set of stamps, not one set of coefficients
+            per detector so this array will have shape
+            (# of stamps, # of zernike coefficients).
```
teweitsai: I do not understand the doc string here. Do you mean the return value is a 2D-matrix/array? In this case, what is the row and what is column? If I am not wrong, the row is the stamps and the column is the Zk. Is this correct? The idea of multiple dimensions matrix/array is that you may want to explain each dimension clearly.

Based on your implementation, you may also want to explain the meaning of `-9999` here or in the `run()` function. I am not sure where is better. Maybe in the `run()`.

python/lsst/ts/wep/task/EstimateZernikesTask.py
```diff
@@ -0,0 +1,335 @@
...
+            exposures[extraExpIdx], donutCatalog, DefocalType.Extra
+        )
+        donutStampsIntra = self.cutOutStamps(
+            exposures[intraExpIdx], donutCatalog, DefocalType.Intra
+        )
+
+        # If no sources in exposure exit with default values
+        if len(donutStampsExtra) == 0:
+            return pipeBase.Struct(
+                outputZernikes=np.ones(19) * -9999,
```
teweitsai: I am also thinking to use the `np.nan`. But this replies on the butler.

doc/uml/taskClass.uml
```diff
@@ -2,4 +2,8 @@
 GenerateDonutCatalogOnlineTaskConfig *-- GenerateDonutCatalogOnlineTaskConnections
 GenerateDonutCatalogOnlineTask *-- GenerateDonutCatalogOnlineTaskConfig
 DonutStamps *-- DonutStamp
+EstimateZernikesTaskConfig *-- EstimateZernikesTaskConnections
+EstimateZernikesTask *-- EstimateZernikesTaskConfig
+EstimateZernikesTask *-- DonutStamps
+EstimateZernikesTask *-- DonutStamp
```
teweitsai: Based on your implementation, the relationship should be:

-------------
EstimateZernikesTaskConfig *-- EstimateZernikesTaskConnections
EstimateZernikesTask *-- EstimateZernikesTaskConfig
EstimateZernikesTask ..> DonutStamps
-------------

`EstimateZernikesTask` makes use of `DonutStamps` in `cutOutStamps()`, but does not hold it or `DonutStamp` as an internal attributes. The composition or aggregation relationship will hold the class as an object in internal attributes.

teweitsai (approved):
This update looks good to me. Please help to correct the `doc/uml/taskClass.uml` as a minimum. In addition, I have some concern for the function to do the cut out. Please consider to improve it or prove my concern is wrong before the merge. Thanks!

doc/uml/taskClass.uml
jbkalmbach: Ok, do I need to include the same for `DonutStamp` since I do hold a list of donutStamp objects inside the function `cutOutStamps`?

python/lsst/ts/wep/Utility.py
jbkalmbach: You're right. That's what I wanted.

python/lsst/ts/wep/task/EstimateZernikesTask.py
jbkalmbach: This is meant to pull in more than one exposure since we are using one exposure for extra-focal and one for intra-focal. Perhaps I should rename this task to `EstimateZernikeFamTask` since we will not want to pull in multiple images when doing this estimation with the wavefront sensors?

jbkalmbach: Then when we are ready for wavefront sensors we will have a different task `EstimateZernikeWfsTask`?

python/lsst/ts/wep/task/EstimateZernikesTask.py
jbkalmbach: In the pipeline yaml file like `testPipeline.yaml`. I will add the configuration defaults to the `testPipeline.yaml` to show where this would happen.

python/lsst/ts/wep/task/EstimateZernikesTask.py
teweitsai: Yes. Please rename it to `EstimateZernikeFamTask`. For the WFS, we can use the `EstimateZernikeWfsTask` as what you said.

python/lsst/ts/wep/task/EstimateZernikesTask.py
teweitsai: Sure. Thanks!

doc/uml/taskClass.uml
teweitsai: No. In your `cutOutStamps()`, it does not have something like the `self.donutStampList`. The function `cutOutStamps()` just uses the `DonutStamps` and `DonutStamp` and returns them. This is the dependency relationship. For the composition or aggregation relationship, they are like this:

Assume you have a class A, and the class B is:

```
class B:
    def __init__(self):
        self.a = A()
```

The class B holds an instance of class A as an internal attribute. Then, depends on the properties of A and B, you might have 

[composition]
-----
B *-- A
-----

or

[aggregation]
-----
B o-- A
-----

python/lsst/ts/wep/task/EstimateZernikesTask.py
jbkalmbach: I will change this to be a padding parameter that adds an extra number of pixels over the `donutStampSize` instead of an initial cutout size. That might be clearer and we will also avoid the situation where initialCutoutSize < donutStampSize.

python/lsst/ts/wep/task/EstimateZernikesTask.py
teweitsai: Maybe we could try the padding for this as well based on your suggestion in the above comments.

jbkalmbach: I think I need a function like you have from before. I can add one here since I will use it for the initial cutout and the final cutout.

python/lsst/ts/wep/task/EstimateZernikesFamTask.py
```diff
@@ -0,0 +1,468 @@
...
+
+        template = templateMaker.makeTemplate(
+            detectorName, defocalType, self.donutTemplateSize
+        )
+
+        return template
+
+    def shiftCenter(self, center, boundary, distance):
+        """Shift the center if its distance to boundary is less than required.
+        Parameters
```
teweitsai, line 153: Please add  a new/empty line before "Parameters".

python/lsst/ts/wep/task/EstimateZernikesFamTask.py
```diff
@@ -0,0 +1,468 @@
...
+        ----------
+        center : float
+            Center point.
+        boundary : float
+            Boundary point.
+        distance : float
+            Required distance.
+        Returns
```
teweitsai, line 162: Please add  a new/empty line before "Returns".

python/lsst/ts/wep/task/EstimateZernikesFamTask.py
```diff
@@ -0,0 +1,468 @@
...
+        # corner of the template
+        templateHalfWidth = int(self.donutTemplateSize / 2)
+        newX = maxLoc[1] - templateHalfWidth
+        newY = maxLoc[0] - templateHalfWidth
+        finalDonutX = xCent + (newX - initialHalfWidth)
+        finalDonutY = yCent + (newY - initialHalfWidth)
+
+        # Shift stamp center if necessary but not final centroid definition
+        xStampCent = copy(finalDonutX)
+        yStampCent = copy(finalDonutY)
```
teweitsai: If `finalDonutX` and `finalDonutY` are `float`, the `copy()` is not needed. You could do some experiment for this.

python/lsst/ts/wep/task/EstimateZernikesFamTask.py
```diff
@@ -0,0 +1,468 @@
...
+        yCornerList = []
+
+        for donutRow in detectorCatalog.to_records():
+            # Make an initial cutout larger than the actual final stamp
+            # so that we can centroid to get the stamp centered exactly
+            # on the donut
+            xCent = int(donutRow["centroid_x"])
+            yCent = int(donutRow["centroid_y"])
+
+            finalDonutX, finalDonutY, xCorner, yCorner = self.calculateFinalCentroid(
```
teweitsai, line 302: Maybe add a comment here to explain why you need to call the `self.calculateFinalCentroid()` for the maintenance or consider to have a more appropriate function name.

teweitsai:
I reviewed the updated and left some comments that you could consider. Thanks!

PR #87, opened 2021-05-11
DM-29961: add FOCUSZ

suberlak:
Add `focusz` to the call to `phosim_repackager.py` in `CreatePhosimDonutTemplates.py`

### DIFF
--- a/python/lsst/ts/wep/CreatePhosimDonutTemplates.py
+++ b/python/lsst/ts/wep/CreatePhosimDonutTemplates.py
@@ -156,16 +156,27 @@ class CreatePhosimDonutTemplates(object):
         runProgram(runPhosimCmd, argstring=runPhosimArgsExtra)
         runProgram(runPhosimCmd, argstring=runPhosimArgsIntra)
 
-    def repackagePhosimImages(self):
+    def repackagePhosimImages(self, defocalDist=1500):
         """
         Run the phosim repackager.
+
+        Parameters
+        ----------
+        defocalDist : int
+            The defocal distance in micrometers, translated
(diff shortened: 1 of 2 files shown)


python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -164,8 +164,12 @@ def repackagePhosimImages(self):
         print("Repackaging phosim output")
 
         argString = f"--out_dir {self.tempWorkPath}/raw "
-        argStringExtra = argString + f"{self.tempWorkPath}/phosimOutput/extra"
-        argStringIntra = argString + f"{self.tempWorkPath}/phosimOutput/intra"
+        argStringExtra = (
+            argString + f"{self.tempWorkPath}/phosimOutput/extra" + " --focusz -1500"
```
teweitsai: Please put the value of `1500` as a function input argument (with the default value of 1500) instead of hard-code. We might need to be able to adjust this value in the commissioning stage.

BTW, do not forget to update the doc string. Thanks!

python/lsst/ts/wep/CreatePhosimDonutTemplates.py
```diff
@@ -164,8 +164,12 @@ def repackagePhosimImages(self):
...
+        )
+        argStringIntra = (
+            argString + f"{self.tempWorkPath}/phosimOutput/intra" + " --focusz +1500"
```
teweitsai: Same comment above. This value is symmetric to the above.

teweitsai (approved):
This update looks good to me. Thanks!

PR #88, opened 2021-05-13
Tickets/dm 30182

jbkalmbach:
We want to have two output tables for the Butler. One with the Zernikes averaged across a chip in a default averaging scheme and one with the complete set of Zernikes for all donut pairs.


teweitsai (approved):
This update looks good to me and very helpful! Thanks for the effort!

PR #89, opened 2021-05-19
Replace `WcsSol` by DM's wcs code in `GenerateDpnutCatalogOnlineTask`.

tribeiro:


doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-1.6.10:
+
+-------------
+1.6.10
+-------------
+
+* Replace ``WcsSol`` by DM's wcs code in ``GenerateDpnutCatalogOnlineTask``.
```
jbkalmbach: Typo ``GenerateDonutCatalogOnlineTask``

doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-1.6.10:
+
+-------------
+1.6.10
```
jbkalmbach: Might want to check with @teweitsai but I think he prefers to roll over to the next minor version number instead of using double digits here (`1.7.0`)

jbkalmbach (approved):
Looks good to me!

python/lsst/ts/wep/task/GenerateDonutCatalogOnlineTask.py
```diff
@@ -176,22 +177,22 @@ def run(
         else:
             raise ValueError(f"{cameraName} is not a valid camera name.")
 
-        # Create WCS holder
-        detWcs = WcsSol(camera=camera)
+        boresightPointing = lsst.geom.SpherePoint(
+            self.boresightRa, self.boresightDec, lsst.geom.degrees
+        )
```
teweitsai, line 182: I think it is better to use the `WcsSol` instead of `lsst.geom.SpherePoint`. The latter one is to use the DM library directly. However, the main reason to have this `WcsSol` is for two reasons:
1. Restrict the way to use the WCS to a single place for the other module to use such as the `ts_phosim`.
2. To prepare for the LSE-75 for the WCS solution provided by DM in the run time. Until now, we do not know the real format yet.

python/lsst/ts/wep/task/EstimateZernikesFamTask.py
```diff
@@ -161,14 +161,14 @@ def assignExtraIntraIdx(self, focusZVal0, focusZVal1):
...
+            extraExpIdx = 1
+            intraExpIdx = 0
         elif focusZVal0 > 0:
             # Check that other image does not have same defocal direction
             if focusZVal1 >= 0:
                 raise ValueError(errorStr)
-            extraExpIdx = 1
-            intraExpIdx = 0
+            extraExpIdx = 0
+            intraExpIdx = 1
```
jbkalmbach, line 171: Looks good. I think this is right based upon the wavefront sensor values in `https://github.com/lsst-ts/ts_wep/blob/develop/policy/focalplanelayout.txt` and the fact that SW0 is extra-focal.

jbkalmbach: Also based upon the coordinate system in the figure in Bo's document here: https://sitcomtn-003.lsst.io/#lsstcam

doc/versionHistory.rst
tribeiro: because of the 10 or the amount of code changes?

teweitsai: Just because of the value of "10". Not the amount of code changes.

python/lsst/ts/wep/task/GenerateDonutCatalogOnlineTask.py
tribeiro: Not sure I understand your point... It is DM that will provide the WCS solution through this very same package.

doc/versionHistory.rst
tribeiro: uhm... I don't think that is really necessary... we adopt semantic versioning (https://semver.org), which means version are subject to the amount of code changes and not incremental.

I will bump up to 1.7 though, because the replacement of WcsSol by DM WCS is probably substantial enough to classify it as minor release. But we should probably be more attentive to this in future releases.

Patch releases are mainly for bug fixes (https://semver.org): 

> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.

And minor releases are for backward compatible functionality:

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.

PR #90, opened 2021-05-22
Tickets/dm 30292

jbkalmbach:
The `append` and `extend` methods in `task/DonutStamps.py` do not work as they are supposed to and we need to update the tests to capture this and add a fix. This fix could go in ts_wep or we could make a change in `meas_algorithms/Stamps.py` by adding a `@property` decorated `stamps` variable like the one there for `metadata`.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,15 @@
 Version History
 ##################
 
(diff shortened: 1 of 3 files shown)


tribeiro (approved):
Looks good.

PR #91, opened 2021-05-25
Update test data with corrected FOCUSZ in headers and update test.

jbkalmbach:
We need to fix the `FOCUSZ` parameter in ts_wep test data for comcam images. The sign of the values in the extra-focal and intra-focal images we have is backwards.


suberlak:
I checked the ticket branch, run `pytest tests/task/test_estimateZernikesFamTask.py`, and it passed. Please update the `versionHistory.rst` and the tag info message with this update. Thanks!

suberlak (commented):

suberlak (approved):

PR #96, opened 2021-06-24
Initial push of donut image design

connolly:
@jbkalmbach can you take a look at the initial design of the image class (merging donutStamp, compensibleImage, and Image)


python/lsst/ts/wep/task/DonutStamp2.py
```diff
@@ -0,0 +1,1473 @@
...
+            # Non-padded mask corresponding to aperture
+            cMask=np.array([], dtype=int),
+            # Coefficient to do the off-axis correction
+            offAxisCoeff=np.array([]),
+            # Defocused offset in files of off-axis correction
+            offAxisOffset=0.0,
+            # True if the image gets the over-compensation
+            caustic=False,
+            centroidFindType=CentroidFindType.RandomWalk,  
+            centroidFind=CentroidFindFactory.createCentroidFind(CentroidFindType.RandomWalk),
```
jbkalmbach, line 132: I think these properties don't need to be a part of the donutStamp. I think we could still have a separate `imageUtils` class or something that handles some things like those in the existing `Image` class.

jbkalmbach (commented):
Started looking this over. As I understand it, we want to turn the DonutStamp class into something that holds not just the cutout as a MaskedImage but also holds the masks and additional information for the wavefront estimate right? That makes sense to me the way you have it. I just think we could try making a utilities class to clean this up a bit. I'll add some more comments as I go through.

PR #97, opened 2021-08-06
Fix donutStamps type error and pipeline error in latest version of stack

jbkalmbach:
This fixes `ts_wep` failing on Jenkins as of August 5. Looks like the error is coming from a change in the stamps.py in `meas_algorithms and missing calibration path in the Gen 3 test pipelines.

### DIFF
--- a/python/lsst/ts/wep/task/DonutStamp.py
+++ b/python/lsst/ts/wep/task/DonutStamp.py
@@ -23,6 +23,8 @@ from dataclasses import dataclass
 
 import lsst.geom
 import lsst.afw.image as afwImage
+import lsst.afw.table as afwTable
+from typing import Optional
 from lsst.meas.algorithms.stamps import AbstractStamp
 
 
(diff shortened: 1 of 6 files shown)


python/lsst/ts/wep/task/DonutStamp.py
```diff
@@ -65,6 +68,8 @@ def factory(cls, stamp_im, metadata, index):
             needed by the constructor.
         idx : int
             Index into the lists in ``metadata``
+        archive_element : `afwTable.io.Persistable`, optional
+            Archive element (e.g. Transform or WCS) associated with this stamp.
```
teweitsai, line 72: Please consider the doc string as:
```
Archive element (e.g. Transform or WCS) associated with this stamp. (the default is None)
```

teweitsai (approved):
The update looks good to me. Thanks for the quick fix of code!

PR #98, opened 2021-08-12
Tickets/DM-30367: update Euler angles

suberlak:
Updating Euler angles in focalplanelayout for SW0 sensors


tests/test_sourceProcessor.py
```diff
@@ -99,14 +99,15 @@ def testCamXYtoFieldXYforWfs(self):
         oxR04SW0, oyR04SW0 = self._camXYtoFieldXY("R04_SW0", 0, 0)
 
         # Compare with the same RXX_SYY
```
teweitsai: We do not compare the same RXX_SYY anymore. Can you update the comment to explain this comparison for the updated Euler angle?

teweitsai (approved):
This update looks good to me. Please help to extend the comments of `test_sourceProcessor.py` to explain the comparison (updated test cases). You could draw some simple figures to explain this comparison to help the future's developer understand the test cases. Thanks!

PR #101, opened 2021-08-19
remove _shiftCenterWfs

suberlak:
Hotfix to DM-30367 removing the shifting wfs centers function, since they're already correct in focalplanelayout


teweitsai (approved):
The update looks good to me. Please help to update the `doc/versionHistory.rst` before the merge. Thanks!

PR #103, opened 2021-08-20
Update to save outputZernikes for pairs of wavefront detectors not ju…

jbkalmbach:
…st a single output for all detectors on focal plane in one.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-1.7.10:
+
+-------------
+1.7.10
+-------------
+
+* Save outputZernikes for pairs of wavefront detectors not just a single output for all detectors.
+
 .. _lsst.ts.wep-1.7.9:
 
 -------------
--- a/python/lsst/ts/wep/task/EstimateZernikesCwfsTask.py
+++ b/python/lsst/ts/wep/task/EstimateZernikesCwfsTask.py
@@ -25,6 +25,7 @@ import pandas as pd
 
 import lsst.pipe.base as pipeBase
 import lsst.afw.image as afwImage
+import lsst.obs.lsst as obs_lsst
 from lsst.pipe.base import connectionTypes
 
 from lsst.ts.wep.Utility import DefocalType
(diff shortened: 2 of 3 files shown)


doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-1.7.10:
+
+-------------
+1.7.10
+-------------
```
teweitsai, line 13: Please consider to use the version `1.8.0` instead. Thanks!

python/lsst/ts/wep/task/EstimateZernikesCwfsTask.py
```diff
@@ -69,6 +97,8 @@ def __init__(self, **kwargs):
         # See LCA-13381 for definition
         self.extraFocalNames = ["R00_SW0", "R04_SW0", "R40_SW0", "R44_SW0"]
         self.intraFocalNames = ["R00_SW1", "R04_SW1", "R40_SW1", "R44_SW1"]
+        self.extraFocalIds = [191, 195, 199, 203]
+        self.intraFocalIds = [192, 196, 200, 204]
```
teweitsai: If we have the `self.extraFocalNames` and `self.intraFocalNames` hard-coded already, do we still need to hard-code the related Ids as well? How could we maintain the hard-coded ID in the future since the detector ID is not fixed in `obs_lsst` in my knowledge? Or the detector ID in `obs_lsst` is fixed already? Is there a way to get the ID programmatically?

teweitsai (approved):
This PR looks good to me. I have some concern for the hard-coded detect IDs. If you could get them programmatically, that will be great. Otherwise, you may want to document them for the future's maintenance.

In addition, if it is possible, please help to write a related unit tests for this update. If this unit test can check the hard-coded detector IDs, that would be great. At least, if the upstream `obs_lsst` changes the behavior, we could catch it immediately. Thanks!

doc/versionHistory.rst
jbkalmbach: @tribeiro mentioned that we want to start using semantic versioning according to the guidelines of https://semver.org/. Since this is a patch level (no new API and backwards compatible) change I think it should increment the lowest values in our version number and go to 1.7.10 instead of 1.8.0.

python/lsst/ts/wep/task/EstimateZernikesCwfsTask.py
jbkalmbach: Ok, we now get the detector mapping directly from `obs_lsst` instead of hard coding them here.

PR #104, opened 2021-09-08
Fix rotation in EstimateZernikesBase. Remove SourceProcessor dependen…

jbkalmbach:
…cy in tasks.


teweitsai (approved):
The update looks good to me. However, the `SourceProcessor` class reads the `EulerZ` angle from the `focalplane.txt` used in `phosim_syseng4` directly. This makes me suspect there might be the inconsistency between `obs_lsst` and `phosim_syseng4`. Please consider to discuss with @suberlak  for the field positions of corner wavefront sensors before the merge.

In addition, it would be great if you could add the unit tests of corner wavefront sensor calculation. Thanks!

jbkalmbach:
What kind of unit tests are you looking for? Should I compare field locations to `focalplane.txt`? I did add this unit test in the previous PR that compares the value I calculate from the pixel locations to the value from the detector orientation information: https://github.com/lsst-ts/ts_wep/blob/develop/tests/task/test_donutStamp.py#L149-L171

jbkalmbach:
And yes, it seems that `obs_lsst` has a 180 degree offset from the `phosim_syseng4` values in the rotation for `R04` and `R40`.

To check the values I can do this:
```
camera = obs_lsst.LsstCam.getCamera()
detector = camera.get('R04_SW0')
detector.getOrientation().getYaw().asDegrees()

270.0
```

teweitsai:
For the unit tests, I am thinking to use the corner wavefront sensor images by PhoSim and calculate the wavefront errors to compare with the truth by OPD. But actually, I think that there might be the problem in this PR and I would like to explain the details in the DM-31636. Thanks!

PR #106, opened 2021-09-28
Remove Unused Code

jbkalmbach:
Remove the old code from ts_wep that we aren't using with the new Gen3 tools.


README.md
```diff
@@ -140,6 +82,49 @@ This requires the configuration file `tests/testData/gen3TestRepo/convertRefCat.
...
+In this case that is found in `tests/testData/pipelineConfigs/testCwfsPipeline.yaml`.
+Looking at this file we see that the three tasks we will run are:
+  - `lsst.ip.isr.isrTask.IsrTask`
+  - `lsst.ts.wep.task.GenerateDonutCatalogOnlineTask.GenerateDonutCatalogOnlineTask`
+  - `lsst.ts.wep.task.EstimateZernikesCwfsTask.EstimateZernikesCwfsTask`
+
+2. Run the `pipetask` from the command line:
+
+```bash
+pipetask run -b $path_of_ts_wep/tests/testData/gen3TestRepo -i refcats/gen2,LSSTCam/raw/all,LSSTCam/calib --instrument lsst.obs.lsst.LsstCam --register-dataset-types --output-run run1 -p $path_of_ts_wep/tests/testData/pipelineConfigs/testCwfsPipeline.yaml -d "exposure IN (4021123106000)"
```
teweitsai, line 105: Please explain the value of `4021123106000` to let the user understand what is this. Thanks!

README.md
```diff
@@ -140,6 +82,49 @@ This requires the configuration file `tests/testData/gen3TestRepo/convertRefCat.
...
+```
+
+The options used above are as follows:
+  - `-b`: Specifies the location of the butler repository.
+  - `-i`: The `collections` (data) needed for the tasks.
+  - `--instrument`: Defines which camera we are using.
+  - `--register-dataset-types`: Add the specific datasets types for our tasks to the registry.
```
teweitsai, line 112: Please give some examples of possible datasets. Thanks!

doc/index.rst
```diff
@@ -18,13 +18,11 @@ Using lsst.ts.wep
...
-* `CamDataCollector` ingests the amplifier images and calibration products based on the DM command line task.
-* `CamIsrWrapper` does the ISR and assemble the CCD images based on the DM command line task.
-* `SourceSelector` queries the bright star catalog (BSC) to select the available target to calculate the wavefront error.
-* `SourceProcessor` processes the post-ISR images to get the clean star images with measured optical field position. The deblending algorithm is used to get the single target bright star image if the neighboring star exists.
 * `WfEstimator` calculates the wavefront error in annular Zernike polynomials up to 22 terms based on the defocal donut images.
-* `WepController` is a high level class to use the WEP package.
+
+Important Pipeline Tasks:
+
+* `EstimateZernikesCwfsTask` calculates the Zernike polynomials on corner wavefront sensors when provided input exposures and source catalog.
```
teweitsai, line 26: This is for the "corner" wavefornt sensors only or not. I thought it could work for the scientific camera (ComCam or LsstFam) as well???

README.md
```diff
@@ -108,21 +65,6 @@ cd $path_of_ts_wep
 setup -k -r .
 ```
 
-## Example Script
-
-- **mapSensorAndFieldIdx.py**: Map the sensor name to the field point index for LSST.
-- **mapSensorAndFieldIdxLsstFam.py**: Map the sensor name to the field point for LSST full-array mode (FAM).
-
```
teweitsai: Please do not remove this section and update these two python scripts to be able to use under this refactored code base. These two files are used to analyze and generate the parameter files in `ts_ofc`. Thanks!.

teweitsai (approved):
This refactor looks good to me. However, I have some questions:

1. Why do you want to remove the `bsc` module? Will you have another PR for this to have the similar functionality? Do not forget, in the real running of telescope, we will need to analyze the bright stars before the exposure images are available.

2. This PR will break the `ts_ofc` (I guess) and `ts_phosim` (I believe). Who will fix them? For the fix of `ts_phosim`, please make sure the closed-loop simulation can still run. I put @suberlak here to let him notice this.

3. Please clean the unused policy files in `policy/` as well. I think the `policy/bsc` might not be needed now. And the `focalplanelayout.txt`. I am not sure the `sensorNameToId.yaml`. For these, maybe you want to check the [ts_MTAOS](https://github.com/lsst-ts/ts_MTAOS). I think this PR might break it as well.

4. Please move the following figure to `README.md`. This is very useful to understand the geometry. You could check with Chris for the details:
https://github.com/lsst-ts/ts_wep/blob/develop/python/lsst/ts/wep/SourceProcessor.py#L170-L186

Thanks!

doc/index.rst
jbkalmbach: I will add the additional task that runs on the LsstFam or ComCam to this list.

doc/index.rst
```diff
@@ -22,6 +22,7 @@ Important classes:
 
 Important Pipeline Tasks:
 
+* `EstimateZernikesFamTask` calculates the Zernike polynomials for LSSTFam or ComCam sensors when provided input exposures and source catalog.
```
teweitsai: The ComCam and FAM are two different "Camera" instances/bodies in real hardware. Please consider to have a better name instead of `EstimateZernikesFamTask` to avoid the confusion.

examples/mapSensorAndFieldIdx.py
```diff
@@ -21,7 +21,58 @@
...
 
-from lsst.ts.wep.SourceProcessor import SourceProcessor
+import lsst.obs.lsst as obs_lsst
+from lsst.afw.cameraGeom import FIELD_ANGLE
+
+
+def getMappingFromFieldXY(fieldXY):
+    """Map the sensor and field index based on the distance between the
+    positions of sensor and field.
+    Parameters
```
teweitsai, line 32: Please add a new line above the `Parameters`. Thanks!

examples/mapSensorAndFieldIdx.py
```diff
@@ -21,7 +21,58 @@
...
+    ----------
+    fieldXY : numpy.ndarray
+        Field (x, y) in degree. This is a nx2 matrix that the first column
+        is the x position and the second column is the y position.
+    Returns
```
teweitsai, line 38: Please add a new line above the `Returns`. Thanks!

README.md
```diff
@@ -122,6 +122,26 @@ The options used above are as follows:
 butler query-collections $path_of_ts_wep/tests/testData/gen3TestRepo/
 ```
 
+## Diagram of the Corner Wavefront Sensor Geometry
+
+    # The wavefront sensors will do the rotation as
+    # following, based on the Euler angle.
```
teweitsai: You could write the following instead directly:
```
The wavefront sensors will do the rotation as following, based on the Euler angle.
```

PR #112, opened 2022-01-08
Tweak test to work with latest meas_algorithms update in w_2022_2.

jbkalmbach:
loadSkyCircle no longer returns centroid column, use loadPixelBox instead.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,15 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.0.2:
+
+-------------
+2.0.2
+-------------
+
+* Patch to work with weekly `w_2022_2`:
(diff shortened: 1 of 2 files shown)


doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.0.2:
+
+-------------
+2.0.2
+-------------
+
+* Patch to work with weekly `w_2022_2`.
```
tribeiro: Can you add the text you entered on the commit here as well, please?

```
loadSkyCircle no longer returns centroid column, use loadPixelBox instead.
```

tribeiro (approved):
Looks good. I think it would be good to add the fix on the version history as well. See my inline comment

PR #113, opened 2022-01-10
Add DonutSourceSelectorTask to task module.

jbkalmbach:


doc/uml/taskClass.uml
```diff
@@ -1,4 +1,5 @@
 @startuml
+class DonutSourceSelectorTask
```
teweitsai: I think it is:
```
DonutSourceSelectorTask *-- DonutSourceSelectorTaskConfig
```

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+
+    ConfigClass = DonutSourceSelectorTaskConfig
+    _DefaultName = "donutSourceSelectorTask"
+
+    def __init__(self, **kwargs):
+        pipeBase.Task.__init__(self, **kwargs)
+
+    def run(self, sourceCat, bbox):
+
```
teweitsai: Please remove this empty line.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+        Returns
+        -------
+        struct : `lsst.pipe.base.Struct`
+            The struct contains the following data:
+            - sourceCat : `lsst.afw.table.SourceCatalog` or `pandas.DataFrame`
+                          or `astropy.table.Table`
+                The catalog of sources that were selected.
+                (may not be memory-contiguous)
+            - selected : `numpy.ndarray` of `bool``
```
teweitsai: There is a "`" in the final, which is not needed. Check the doc of "bool"

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+        struct : `lsst.pipe.base.Struct`
+            The struct contains the following data:
+            - selected : `numpy.ndarray` of `bool``
+                Boolean array of sources that were selected, same length as
+                sourceCat.
+
+        Raises
+        ------
+        ValueError
+            sourceLimit in config for task must be -1 or a positive integer
```
teweitsai: Please add a period in the end of sentence.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+        if self.config.sourceLimit > 0:
+            doSourceLimit = True
+        elif self.config.sourceLimit == -1:
+            doSourceLimit = False
+        else:
+            errMsg = str(
+                "config.sourceLimit must be a positive integer "
+                + "or turned off by setting it to '-1'"
+            )
+            raise ValueError(errMsg)
```
teweitsai: I think you could put this `if` case to the initial beginning of this function to avoid the calculation above if you will raise the `ValueError` instead. In addition, you could get rid off the argument of `doSourceLimit` actually
.

I mean,
```
if not ((self.config.sourceLimit == -1) or (self.config.sourceLimit > 0)):
    raise ValueError("")
```

teweitsai: In addition, the original or above code does not check the `self.config.sourceLimit` is an integer or not. You could do `self.config.sourceLimit = int(self.config.sourceLimit)` or add more condition to check integer if you really need the user to put a positive integer value.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+                elif (maxBlended == 0) and (np.min(magDiff) >= minMagDiff):
+                    index.append(groupIndices[srcOn])
+                    sourcesKept += 1
+                elif len(magDiff) <= maxBlended:
+                    index.append(groupIndices[srcOn])
+                    sourcesKept += 1
+                else:
+                    if np.partition(magDiff, maxBlended)[maxBlended - 1] > minMagDiff:
+                        index.append(groupIndices[srcOn])
+                        sourcesKept += 1
```
teweitsai: The code here has three layers of indentation (`if` case). Please consider to simplify the code if possible. The idea is to try to decease the layers of indentation if possible. Thanks!

I also tried to understand this part of code here. It is not very clear to me. Please consider to refactor it to be more clear if possible.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -0,0 +1,225 @@
...
+
+            if doSourceLimit and (sourcesKept == self.config.sourceLimit):
+                break
```
teweitsai: As the comment above, I felt the argument of `doSourceLimit` may not be needed. In this case, it might be:

```
if (self.config.sourceLimit > 0) and (sourcesKept == self.config.sourceLimit):
    break
````

File: tests/task/test_donutSourceSelectorTask.py [hunk 8]
```diff
@@ -0,0 +1,142 @@
...
+        minimalCat["centroid_y"] = [100.0, 100.0, 100.0, 100.0]
+        bbox = Box2I(Point2I(0, 0), Point2I(500, 500))
+        return minimalCat, bbox
+
+    def testValidateConfigs(self):
+
+        self.config.xCoordField = "center_x"
+        self.task = DonutSourceSelectorTask(config=self.config, name="New Task")
+
+        self.assertEqual(self.task.config.xCoordField, "center_x")
```
teweitsai: Why it is `center_x` instead of `centroid_x` here? And why it can pass???

tests/task/test_donutSourceSelectorTask.py
```diff
@@ -0,0 +1,142 @@
...
+        self.config.sourceLimit = -1
+        self.task = DonutSourceSelectorTask(config=self.config, name="Test Task")
+        testCatStruct = self.task.selectSources(minimalCat, bbox)
+        testCatSelected = testCatStruct.selected
+        self.assertEqual(len(testCatSelected), 4)
+        self.assertEqual(sum(testCatSelected), 2)
+
+        # Test that only the brightest blended on the end that only
+        # blends with one other donut is accepted even though
+        # it does not meet is isoMagDiff criterion we allow it because
```
teweitsai: "it does not meet is isoMagDiff criterion" or "it does not meet the isoMagDiff criterion"?

teweitsai (approved):
This PR looks good to me. Please consider the comments before the merge. Thanks!

tests/task/test_donutSourceSelectorTask.py
jbkalmbach: Here I was checking that if you change the configuration (in line 59) in the case that your input catalog has columns with different names that this configuration will be passed along. I can add a test here to make sure that it will fail if this configuration does not change before we change it.

teweitsai: Sure. Please do so. Thanks!

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
jbkalmbach: This should be handled by the configuration class. In that class we define the field to be integer only. I will add a test to make sure that an error is raised if one tries to set the configuration value with the wrong type.

PR #117, opened 2022-01-26
Fix donut source selector maxBlended parameter.

jbkalmbach:


teweitsai (approved):
Looks good. Thanks for this fix of bug.

PR #124, opened 2022-02-28
DM-32023: Use GalSim for Zernike and Cartesian polynomial evaluation

jmeyers314:
This effectively fixes a bug in `ZernikeAnnularJacobian` when `order="2nd"`.  

It should also be faster, at least when vectorized inputs are used.  `lsst-devl03` is not a stable environment, so this may be spurious, but my initial timing test revealed that 
tests/task/test_estimateZernikesCwfsTask.py went from 364s to 129s.


jmeyers314:
Used wrong ticket number.  Closing...

PR #125, opened 2022-03-18
Update Jenkinsfile to always pull the image before new builds and imp…

tribeiro:
…rove cleanup stages to make build more robust.


teweitsai (approved):

PR #126, opened 2022-03-18
DM-33169: Use GalSim for Zernike/Cartesian polynomial evaluation

jmeyers314:
GalSim is faster than the code it's replacing (at least for vectorized inputs).

Also fixes an error in `ZernikeAnnularJacobian` when `order=2` (also known as the Hessian determinant).  The `mathcwfs` algorithm incorrectly assumes the Hessian is linear in the Zernike coefficients, and thus you can just add up the contribution from each index independently.  Since the Hessian involves products of polynomials, however, it's not linear in the coefficients.  GalSim performs this operation correctly by correctly implementing the multiplication.

Finally, this PR also makes this package python only.  Yay!


python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -853,8 +859,24 @@ def _poly10Grad(self, c, x, y, atype):
         numpy.ndarray
             Corrected parameters for off-axis distortion.
         """
-
-        return mathcwfs.poly10Grad(c, x.flatten(), y.flatten(), atype).reshape(x.shape)
+        if atype not in ["dx", "dy"]:
```
jbkalmbach, line 867: Can you add information on this to the docstring under a:
```
Raises
-------
```
section?

python/lsst/ts/wep/cwfs/Tool.py
```diff
@@ -110,27 +71,42 @@ def ZernikeAnnularGrad(z, x, y, e, axis, nMax=22):
...
+        return gszk.gradX(x, y)
+    elif axis == "dy":
+        return gszk.gradY(x, y)
+    elif axis == "dx2":
+        return gszk.gradX.gradX(x, y)
+    elif axis == "dy2":
+        return gszk.gradY.gradY(x, y)
+    elif axis == "dxy":
+        return gszk.gradX.gradY(x, y)
+    else:
```
jbkalmbach, line 104: Once again here can you add information on the error to the docstring?

python/lsst/ts/wep/cwfs/Tool.py
```diff
@@ -142,26 +118,31 @@ def ZernikeAnnularJacobian(z, x, y, e, order, nMax=22):
...
-    # Calculate the Jacobian
-    return mathcwfs.zernikeAnnularJacobian(
-        z, x.flatten(), y.flatten(), e, order
-    ).reshape(x.shape)
+    gszk = GSZernike(np.concatenate([[0], z]), R_inner=e)
+    if order == "1st":
+        return gszk.laplacian(x, y)
+    elif order == "2nd":
+        return gszk.hessian(x, y)
+    else:
```
jbkalmbach, line 151: Same comment as above.

python/lsst/ts/wep/cwfs/Tool.py
```diff
@@ -38,64 +39,24 @@ def ZernikeAnnularEval(z, x, y, e, nMax=28):
         Y coordinate on pupil plane.
     e : float
         Obscuration value. It is 0.61 in LSST.
-    nMax : int, optional
-        Maximum number of Zernike terms. (the default is 28.)
+    nMax : int, deprecated
```
jbkalmbach, line 42: Just curious. How many does Galsim return?

doc/versionHistory.rst
```diff
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.2.5:
```
jbkalmbach: Might have to coordinate with @suberlak on the order you two choose to merge your PRs and adjust this file accordingly.

jbkalmbach (approved):
Please add galsim to the required packages on the repository README. 

Other than that looks good to me just some additional inline documentation comments. Happy to have this python only!

doc/versionHistory.rst
suberlak: I'm sorry I saw this comment after I merged; perhaps 2.3.1 ?

python/lsst/ts/wep/cwfs/Tool.py
jmeyers314: GalSim generates the Zernike polynomials on the fly programmatically, so in principle there's no limit.  That said, I'm sure the recurrence relations used accrue floating-point errors at some rate.  I'd trust Z100, and maybe even Z1000, but probably not Z10000.  Most of the GalSim unit tests for zernikes top out around Z30, but I know it's accurate well beyond this.

doc/versionHistory.rst
jmeyers314: np.  I'll update the revision # as needed.

jbkalmbach:
Oh, just thought of something else. You should probably remove the C++ lines from the `.githooks/pre-commit` file as well.

jmeyers314:
> Oh, just thought of something else. You should probably remove the C++ lines from the .githooks/pre-commit file as well.

Should we just leave them maybe?  I don't think they're doing any harm.  And it'd be easy to forget to re-add them if we ever add c++ code here again.

jbkalmbach:
> > Oh, just thought of something else. You should probably remove the C++ lines from the .githooks/pre-commit file as well.
> 
> Should we just leave them maybe? I don't think they're doing any harm. And it'd be easy to forget to re-add them if we ever add c++ code here again.

Sure, if it's not bothering anything. I thought maybe installing `clang-format` is an extra step for some people but if it's commonly part of operating systems then I guess it doesn't really do any harm to leave it.

jmeyers314:
> I thought maybe installing clang-format is an extra step for some people

Looks like that part of the hook doesn't even trigger if `find` finds no .cc or .h files, so should be okay.

PR #127, opened 2022-03-25
DM-34187: make auxTel test more robust

suberlak:


jmeyers314 (commented):
Hi Chris,

This looks okay to me, except that I appear to gain a new collection every time I run `scons` now, which doesn't seem good.  For example, if you run `butler query-collections /repo/main "u/jemeyers/*"` now, you'll see 5 tempfile collections.  It'd be nice if these got deleted when the test passes.

suberlak:
I added a `tearDownClass` which does the clean up - when I run it now I don't see any temporary runs in the `/repo/main/$USER` . Rebased against develop,  updated version history against #128.

jmeyers314 (approved):
Thanks Chris,
LGTM.

PR #128, opened 2022-03-29
Adjust CombineZernikesSigmaClipTask methods.

jbkalmbach:


tests/task/test_combineZernikesBase.py
```diff
@@ -45,6 +46,7 @@ def combineZernikes(self, zernikeArray):
         self.assertEqual(type(taskOutput), pipeBase.Struct)
         np.testing.assert_array_equal(taskOutput.combinedZernikes, np.arange(10))
         np.testing.assert_array_equal(taskOutput.flags, np.ones(10))
+        self.assertTrue(isinstance(taskOutput.flags[0], numbers.Integral))
```
teweitsai, line 49: I have a question. Is this new test is to check the `taskOutput.flags[0]` is `int` or not? If so, why not use the `type(taskOutput.flags[0])` and check the output is `int` or not? This can remove the dependency of `numbers` module? Or did I misunderstand the purpose here? I am curious because I noticed you put this checking in several test scripts in this PR. Thanks!

teweitsai (approved):
Looks good to me. Thanks for this improvement!

tests/task/test_combineZernikesBase.py
jbkalmbach: When I was testing this out when I was coming from a numpy array `type(taskOutput.flags[0]) == int` I always got `False` because the integers in the numpy array are `numpy.int64`. But if I use the base class from `numbers.Integral` then I get `True` no matter if it is a `numpy.int` or basic python `int`.

teweitsai: I see your point now. Thanks!

PR #132, opened 2022-04-26
Fix rotation of sensors in EstimateZernikesBase.

jbkalmbach:
This fixed a problem that was preventing the closed loop from converging with the CWFS:

10 iterations before change:
image

After change:
fwhmIters

### DIFF
--- a/python/lsst/ts/wep/task/EstimateZernikesBase.py
+++ b/python/lsst/ts/wep/task/EstimateZernikesBase.py
@@ -482,8 +482,8 @@ class EstimateZernikesBaseTask(pipeBase.PipelineTask):
             detectorExtra = camera.get(donutExtra.detector_name)
             detectorIntra = camera.get(donutIntra.detector_name)
 
-            eulerZExtra = detectorExtra.getOrientation().getYaw().asDegrees()
-            eulerZIntra = detectorIntra.getOrientation().getYaw().asDegrees()
+            eulerZExtra = -1.0 * detectorExtra.getOrientation().getYaw().asDegrees()
+            eulerZIntra = -1.0 * detectorIntra.getOrientation().getYaw().asDegrees()
 
             # NOTE: TS_WEP expects these images to be transposed
             # TODO: Look into this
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/task/EstimateZernikesBase.py
```diff
@@ -482,8 +482,8 @@ def estimateZernikes(self, donutStampsExtra, donutStampsIntra):
             detectorExtra = camera.get(donutExtra.detector_name)
             detectorIntra = camera.get(donutIntra.detector_name)
 
-            eulerZExtra = detectorExtra.getOrientation().getYaw().asDegrees()
-            eulerZIntra = detectorIntra.getOrientation().getYaw().asDegrees()
+            eulerZExtra = -1.0 * detectorExtra.getOrientation().getYaw().asDegrees()
+            eulerZIntra = -1.0 * detectorIntra.getOrientation().getYaw().asDegrees()
```
teweitsai: Please help to comment the reason for this `-1.0`. In addition, maybe:

```
eulerZExtra = -detectorExtra.getOrientation().getYaw().asDegrees()
eulerZIntra = -detectorIntra.getOrientation().getYaw().asDegrees()
```

Is it possible to add the unit test for this change?

teweitsai (approved):
Looks good. Thanks for the fix of code!

python/lsst/ts/wep/task/EstimateZernikesBase.py
teweitsai: BTW, will this fix affect the simulation of ComCam or LSST FAM? I guess no because the Euler angle of scientific camera CCDs is ~0. But just want to confirm. Thanks!

teweitsai: In addition, is it possible that the problem comes from the upstream `obs_lsst` or the downstream `ts_phosim` (`focalplane.txt`)? Maybe @suberlak could comment. Thanks!

jbkalmbach: > 

That's correct. This will not affect ComCam or LSST FAM because the Euler angle for all of those detectors is ~0.

jbkalmbach: I couldn't think of an exact way to test the rotation but I did find that the results we get are better matched in the test than before so we can tighten the allowed error between the OPD truth and the task results for the corner wavefront sensors.

PR #133, opened 2022-04-30
DM-34614: Remove phosim_utils dependency

jmeyers314:
Remove phosim_utils dependency


tribeiro (approved):
Looks good! setup dependency removed as well as setting it up on CI job and tests still passes, so I think we're good.

PR #135, opened 2022-05-19
DM-34879: Don't switch matplotlib backend on import

jmeyers314:


jmeyers314:
Note, I'm not sure the previous code could even run if `saveFilePath = None`, since `plt.show()` with a non-interactive backend will fail I think.   

I could be convinced to just remove this file entirely, since it's doing mostly trivial things that could easily be reproduced without a dedicated plotting function.  But for the moment, I just want to remove the switch-backend-on-import.

teweitsai (approved):
This PR looks good to me!

Please help to update the `doc/versionHistory.rst` for this.

I think the function of `plotZernike()` will fail in some cases (this is why I added that `plt.switch_backend()` before. You could remove this `PlotUtil.py` if you think it is useless. 

Thanks!

jmeyers314:
Thanks @teweitsai .  I updated the history and will tag a new release after the merge.  I decided to keep the file around, just in case anyone finds it useful.  And I also don't need to worry about going through a deprecation period this way.

PR #139, opened 2022-06-14
Tickets/dm 34846

suberlak:
Updates mostly in CompensableImage, from cMask to mask_comp, and pMask to mask_pupil. However, in some places in Algorithm.py, cMask was meant as the non-padded  (pupil) mask, and pMask as the padded (computational) mask, so it was updated accordingly.

### DIFF
--- a/tests/cwfs/test_multiImgs.py
+++ b/tests/cwfs/test_multiImgs.py
@@ -43,8 +43,8 @@ class TestWepWithMultiImgs(BaseCwfsTestCase, unittest.TestCase):
         )
 
         # Set the tolerance
-        self.tolMax = 4
-        self.tolRms = 1
+        self.tolMax = 8.5
+        self.tolRms = 3
 
         # Record the start time
         self.startTime = time.perf_counter()
--- a/tests/test_wfEstimator.py
+++ b/tests/test_wfEstimator.py
@@ -27,7 +27,7 @@ from lsst.ts.wep.WfEstimator import WfEstimator
 from lsst.ts.wep.Utility import getModulePath, getConfigDir, DefocalType, CamType
 
 
-class TestWfEsitmator(unittest.TestCase):
+class TestWfEstimator(unittest.TestCase):
     """Test the wavefront estimator class."""
 
     def setUp(self):
@@ -141,7 +141,7 @@ class TestWfEsitmator(unittest.TestCase):
         ]
         zer4UpNm = self.wfsEst.calWfsErr()
         self.assertAlmostEqual(
-            np.sum(np.abs(zer4UpNm - np.array(wfsError))), 6.95092306, places=7
+            np.sum(np.abs(zer4UpNm - np.array(wfsError))), 28.38981929, places=7
         )
 
         # Test to reset the data
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.5.0:
+
+-------------
+2.5.0
+-------------
+
+* Update cMask, pMask to mask_pupil, mask_comp.
+
 .. _lsst.ts.wep-2.4.4:
 
 -------------
--- a/python/lsst/ts/wep/cwfs/Algorithm.py
+++ b/python/lsst/ts/wep/cwfs/Algorithm.py
@@ -96,14 +96,14 @@ class Algorithm(object):
         self.zc = np.array([])
 
         # Padded mask for use at the offset planes
-        self.pMask = None
+        self.mask_comp = None
 
         # Non-padded mask corresponding to aperture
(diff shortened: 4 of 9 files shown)


tests/test_wfEstimator.py
```diff
@@ -141,7 +141,7 @@ def testCalWfsErrOfFft(self):
         ]
         zer4UpNm = self.wfsEst.calWfsErr()
         self.assertAlmostEqual(
-            np.sum(np.abs(zer4UpNm - np.array(wfsError))), 6.95092306, places=7
+            np.sum(np.abs(zer4UpNm - np.array(wfsError))), 28.38981929, places=7
```
teweitsai, line 144: It is weird to me that why the rename of variables will affect the calculation result and you need to modify the tolerance to pass the unit test. Is this because of the change of `numpy` version by accident?

tests/cwfs/test_multiImgs.py
```diff
@@ -43,8 +43,8 @@ def setUp(self):
         )
 
         # Set the tolerance
-        self.tolMax = 4
-        self.tolRms = 1
+        self.tolMax = 8.5
+        self.tolRms = 3
```
teweitsai, line 47: It is weird to me that why the rename of variables will affect the calculation result and you need to modify the tolerance to pass the unit test. Is this because of the change of `numpy` version by accident?

teweitsai (approved):
This PR looks good to me. However, I am a little concerned that there is the change of tolerance to pass the unit test. I am not sure this is because of the change of `numpy` version by accident or not. I assumed the change of variable names should not affect the calculation result.

When I checked the `CompensableImage.py`, I noticed that some `self.pMask` was renamed to `self.mask_comp` and some was renamed to `self.mask_pupil`. But for those part, I do not see the affection of calculation. I reviewed the PR three times to make sure there should be no calculation process change.

Please help to double check this PR and merge it if you do not feel anything wrong. But please consider to document the `versionHistory.rst` to say there is "something" change and you need to modify the tolerance value. If you could figure out what is that "something" change, it will be great.

Thanks!

tests/test_wfEstimator.py
suberlak: This is because previously incorrect masks were returned by `CompensableImage` methods of `getPaddedMask` (which returned `pMask`) and `getNonPaddedMask` (which returned `cMask`) https://github.com/lsst-ts/ts_wep/blob/92110cfdaaacfcef9a9821cf4cb920be2c7d9487/python/lsst/ts/wep/cwfs/CompensableImage.py#L172 . In fact, `CompensableImage` calculates masks  in a way that `cMask` is actually `mask_comp` i.e. "padded", whereas `pMask` is actually `mask_pupil`, i.e. "non-padded". Thus notice that  the methods above needed to return different masks (i.e. `getPaddedMask` returns `mask_comp` and `getNonPaddedMask` returns `mask_pupil` ). Before whenever `Algorithm.py` asked for a "non-padded" mask it received a `cMask`, which was actually padded. So it got the wrong mask from `CompensableImage`. The AOS converges as before ( as I show in the notebook in the ticket comments), but I did expect small differences since now the correct masks are used (i.e. when `Algorithm.py` asks for `getNonPaddedMasks` it receives the `mask_pupil` which is indeed non-padded https://github.com/lsst-ts/ts_wep/blob/d41a81fe380c09432950d0db933e7b664b8782ea/python/lsst/ts/wep/cwfs/Algorithm.py#L1181 ).

tests/cwfs/test_multiImgs.py
suberlak: If this comparison was to Bo's cwfs code, which also got the masks wrong, then I would not be surprised. It may be that `cMask` and `pMask` were unnoticed in his old code too (eg. `pMask` calculated as "pupil-mask", but treated as "padded-mask", which it isn't, and `cMask` calculated as "computational-mask" but treated as "non-padded", which it isn't).

suberlak: What I meant is that it may be that what is "truth" would have to be revised here if "truth" used wrong masks.

tests/test_wfEstimator.py
teweitsai: Sure. Thanks for the explanation!

tests/cwfs/test_multiImgs.py
teweitsai: Sure. Thanks for the explanation!

suberlak: I quickly checked and the main differences are where the `test_multImgs` uses `fft` , i.e. cases 2, 3, and 4: 
image

image

image

this makes sense because `fft` uses the padded mask (from  `boundaryT` docstring " in fft, it is also the width of Neuman boundary where the derivative of the wavefront is set to zero.")

PR #141, opened 2022-06-17
Tickets/dm 34846: revert some changes

suberlak:
It turns out that the algorithm worked with pMask having the non-padded mask, and cMask having the padded mask, despite them being inherited via 

     self.pMask = I1.getPaddedMask() * I2.getPaddedMask()

but since previously `getPaddedMask` returned `pMask` which was `non-padded`,  it ended up with `pMask` working as `mask_pupil` despite  being inherited from `getPaddedMask`.

Thus now just like in `Compensable` , `pMask` --> `mask_pupil` ( "non-padded"), and `cMask` --> `mask_comp` ("padded"), and the what makes it consistent is that `mask_pupil` is inherited from `getNonPaddedMask` (which returns indeed `mask_pupil`.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -12,7 +12,8 @@ Version History
 2.5.0
 -------------
 
-* Update cMask, pMask to mask_pupil, mask_comp.
+* Update names of cMask to mask_comp (padded), pMask to mask_pupil (non-padded)
+* Correct output of getPaddedMask to mask_comp, getNonPaddedMask to mask_pupil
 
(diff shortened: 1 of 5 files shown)


teweitsai (approved):
Looks good to me. Thanks for the quick fix! Please help to update the version history as well. Or retag it. Either way works to me. Thanks!

suberlak:
In short, now the change of PR139 & PR141 combined does:

cMask --> mask_comp
pMask --> mask_pupil

(in both `compensableImage` and `Algorithm`)

PR #142, opened 2022-06-24
Fixed docstring in Algorithm.py to indicate that I1 must be intrafocal and I2 must be extrafocal.

jfcrenshaw:
I thought about renaming I1 and I2 to more descriptive names (e.g. intraImg, extraImg), but didn't want to introduce backwards-incompatibility. Let me know if you think that change would be preferred.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.5.1:
+
+-------------
+2.5.1
+-------------
+
+* Fix docstrings in Algorithm.py to indicate that I1 must be intrafocal and I2 must be extrafocal.
+
 .. _lsst.ts.wep-2.5.0:
 
 -------------
--- a/python/lsst/ts/wep/cwfs/Algorithm.py
+++ b/python/lsst/ts/wep/cwfs/Algorithm.py
@@ -518,9 +518,9 @@ class Algorithm(object):
         Parameters
         ----------
         I1 : CompensableImage
-            Intra- or extra-focal image.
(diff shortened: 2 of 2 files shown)


teweitsai:
For the variable names, I prefer to put the common parts first such as `imgIntra` and `imgExtra`. By doing this, when you are using the IDE or Jupiter notebook, you can use the `tab`+`tab` (or other shortcut) to find the available terms more quickly.

jfcrenshaw:
That makes sense. Are you saying you want me to change `I1` and `I2` to `imgIntra` and `imgExtra`?

teweitsai:
No. The `I1` + `I2` or `imgIntra` + `imgExtra` works to me. I just mean the naming convention of variables. Sorry to mislead you. Thanks!

jbkalmbach:
Looks fine to me. If you do want to change to `imgIntra` and `imgExtra` I would support that. I don't think it will break anything since it seems these `I1` and `I2` calls are all limited to within the functions of `Algorithm.py`.

You'll need to rebase and update your `doc/versionHistory.rst`.

jbkalmbach (approved):
Looks fine to me. If you do want to change to imgIntra and imgExtra I would support that. I don't think it will break anything since it seems these I1 and I2 calls are all limited to within the functions of Algorithm.py.

You'll need to rebase and update your doc/versionHistory.rst.

jfcrenshaw:
Rather than changing this documentation, I believe the I1/I2 symmetry can be restored by fixing the mask rotations in the algorithm. Once that gets sorted out, I will check if the symmetry is restored, and either close or merge this PR, depending on the results of that test.

PR #143, opened 2022-07-01
Add comcam to donutTemplateModel.

jbkalmbach:


suberlak (commented):
I guess it hasn't really been in test coverage before, but shouldn't perhaps testing for allowed `camType` be part of https://github.com/lsst-ts/ts_wep/blob/512400f6fd9f4883b99638ef120bfab8418ee25e/tests/cwfs/test_donutTemplateModel.py#L55 `testMakeTemplate` ? Something that would test the allowed values and that the exception is raised if `camType` is outside of allowed values

suberlak (approved):

jbkalmbach:
> I guess it hasn't really been in test coverage before, but shouldn't perhaps testing for allowed `camType` be part of
> 
> https://github.com/lsst-ts/ts_wep/blob/512400f6fd9f4883b99638ef120bfab8418ee25e/tests/cwfs/test_donutTemplateModel.py#L55
> 
> `testMakeTemplate` ? Something that would test the allowed values and that the exception is raised if `camType` is outside of allowed values

Added a test for this along with a new error message specifying that running AuxTel only works with the `onAxis` optical model and an accompanying test.

PR #144, opened 2022-07-11
Update gen3TestRepo to be compatible with latest version of stack (w_…

jbkalmbach:
…2022_28).


tests/task/test_calcZernikesTaskCwfs.py
```diff
@@ -63,7 +63,7 @@ def setUpClass(cls):
             cleanUpCmd = writeCleanUpRepoCmd(cls.repoDir, cls.runName)
             runProgram(cleanUpCmd)
 
-        collections = "refcats,LSSTCam/calib,LSSTCam/raw/all"
+        collections = "refcats/gen2,LSSTCam/calib,LSSTCam/raw/all"
```
suberlak, line 66: does that mean that we are still using legacy gen2? Is there a way to do the refcat generation here only in Gen3 ?

suberlak (changes requested):
I wonder about the gen2 refcats. Are we still relying on old gen2 refcat generation? Is it possible to use only gen3 at this point ?

tests/task/test_calcZernikesTaskCwfs.py
jbkalmbach: There's no refcat generation here. They are already generated. They are just labelled this way because I originally used gen2->gen3 conversion to create them when I first added the data into the repository.

suberlak (approved):

tests/task/test_calcZernikesTaskCwfs.py
suberlak: Thanks for clarification! Looks good to merge.

PR #145, opened 2022-08-06
DM-35860: Update science sensor tasks to get focusZ from exposure visitInfo

jbkalmbach:
Update science sensor tasks to get focusZ from exposure visitInfo instead of metadata after update in DM-35186.

### DIFF
--- a/python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py
+++ b/python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py
@@ -142,9 +142,11 @@ class EstimateZernikesScienceSensorTask(EstimateZernikesBaseTask):
         camera: lsst.afw.cameraGeom.Camera,
     ) -> pipeBase.Struct:
 
-        # Get exposure metadata to find which is extra and intra
-        focusZ0 = exposures[0].getMetadata()["FOCUSZ"]
-        focusZ1 = exposures[1].getMetadata()["FOCUSZ"]
+        # Get exposure visitInfo to find which is extra and intra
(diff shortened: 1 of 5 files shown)


python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py
```diff
@@ -142,9 +142,11 @@ def run(
         camera: lsst.afw.cameraGeom.Camera,
     ) -> pipeBase.Struct:
 
-        # Get exposure metadata to find which is extra and intra
```
suberlak, line 145: Perhaps it would be useful to throw an exception here if such a method is not available (meaning the data has not been run through 'defineVisits' with the post-DM-35186 butler)? If I understand correctly data properly ingested in gen3 repos earlier have "FOCUSZ" keyword header, but not `visitInfo.focusZ` property. Perhaps if `focusZ` is not available, it would be good to catch that exception here.

suberlak (commented):
Looks good! Will it work with only with all data ingested post-35186, or is there any scenario that this change will break if trying to work with older (gen3) data?

suberlak (approved):

python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py
jbkalmbach: Just tested on the LATISS data from `'u/scichris/Latiss/postISRtest'`. Looks like everything works as long as focusZ was in the original header. After update to `EstimateZernikesLatissTask` added I successfully ran `tests/task/test_estimateZernikesLatissTask.py ` as well.

PR #147, opened 2022-09-08
Fix task input order in test_estimateZernikes... tests.

jbkalmbach:


suberlak (commented):
Thus far there is a difference which image is which when passed on to `Algorithm.py` due to rotations that are only done on one of them. Have you checked that the same result is obtained whether extra/intra is provided rather than intra/extra ? (I think JFC's update would fix that https://github.com/lsst-ts/ts_wep/pull/142 but it's not yet merged )

jbkalmbach:
The tasks as they are written always run the same way no matter which order you pass to the `run` command because they all have functions like the following:

https://github.com/lsst-ts/ts_wep/blob/7409c05e1e9ba236f3a3e8bfb8b503246ca5d72a/python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py#L92

that figure out which exposure is which and then pass them to the `estimateZernikes` function the same way every time.  So if there is an issue with if the order in `estimateZernikes` makes a difference that should be a separate issue.

suberlak (approved):

suberlak:
Ok, thanks for the explanation!

PR #148, opened 2022-09-15
tickets/DM-36253

jbkalmbach:
Change focusZ in headers of repackaged phosim data to be in mm instead of microns after phosim_utils update.


suberlak (approved):
I can't really see the change in the binary files, but if all that's changed is the header content then all should work well, thanks for the update!

PR #152, opened 2022-09-28
Tickets/DM-36154: Algorithm history; symmetry with respect to I1, I2.

jfcrenshaw:
### Algorithm Log:
This PR adds internal logging to the algorithm, which you can see in action in [this notebook](https://gist.github.com/jfcrenshaw/751848194c27a1e79b6f51d20193a48c).

There is logging for the outer and inner loops of the algorithm, including the initial images, the compensated images, the masks, the estimated zernikes and wavefronts, and the reason for terminating the outer loop.

You can set the `debugLevel` at which each object is logged. I have set everything to be logged when `debugLevel>=1`, but we can change this if people want.

*The algorithm log is not saved in the butler when running the pipeline tasks. I will open a separate ticket aimed at getting the log integrated in the pipeline tasks.*

### Other changes:

In the process of setting up this logging, I restored the symmetry of the algorithm with respect to `I1` and `I2`. This required two things:

#### 1. Simplifying rotations of the masks and images

After this update, masks match the orientations of the original images (i.e. the orientation of the extrafocal and intrafocal masks are different by 180 deg, because the orientation of the corresponding images are different by 180 deg).

The orientation of the extrafocal image is made to match the orientation of the intrafocal image during image compensation, which happens while running the algorithm. The algorithm then applies masks to the compensated images. For this reason, when the algorithm creates the masks, it calls `makeMask(..., compensated=True)`. This causes the masks to match the orientation of the compensated images (i.e. the extrafocal mask is now rotated 180 deg to match the orientation of the compensated extrafocal image).

#### 2. Simplified and fixed the calculation of `I0` and `dI`

The algorithm requires the calculation of the mean image, `I0`, and the difference image, `dI`. The "exp" and "fft" algorithms were both doing this, using separate bits of code. I removed the code duplication, so both algorithm modes are using the same method to calculate these images (i.e. `Algorithm._getdIandI0`).

The sign of the difference image, `dI`, is sensitive to which image is extrafocal and which image is intrafocal. I fixed this by introducing an additional minus sign whenever `I2` is extrafocal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 This module calculates the wavefront error in annular Zernike polynomials based on the intra- and extra-focal donut images in the large synoptic survey telescope (LSST).
 
-See the docs: https://ts-wep.lsst.io/
+See the docs: <https://ts-wep.lsst.io/>
 
 ## Platform
 
@@ -97,9 +97,10 @@ Here we show how to run the pipeline on the LSSTCam corner wavefront sensors.
 1. The order of the tasks and the configuration overrides for the tasks are set in the pipeline definition `yaml` file.
 In this case that is found in `tests/testData/pipelineConfigs/testCwfsPipeline.yaml`.
 Looking at this file we see that the three tasks we will run are:
-  - `lsst.ip.isr.isrTask.IsrTask`
-  - `lsst.ts.wep.task.GenerateDonutCatalogOnlineTask.GenerateDonutCatalogOnlineTask`
-  - `lsst.ts.wep.task.EstimateZernikesCwfsTask.EstimateZernikesCwfsTask`
+
+- `lsst.ip.isr.isrTask.IsrTask`
+- `lsst.ts.wep.task.GenerateDonutCatalogOnlineTask.GenerateDonutCatalogOnlineTask`
+- `lsst.ts.wep.task.EstimateZernikesCwfsTask.EstimateZernikesCwfsTask`
 
 2. Run the `pipetask` from the command line:
 
@@ -108,15 +109,16 @@ pipetask run -b $path_of_ts_wep/tests/testData/gen3TestRepo -i refcats/gen2,LSST
 ```
 
 The options used above are as follows:
-  - `-b`: Specifies the location of the butler repository.
-  - `-i`: The `collections` (data) needed for the tasks.
-  - `--instrument`: Defines which camera we are using.
-  - `--register-dataset-types`: Add the specific datasets types for our tasks to the registry.
-    - Dataset Types that are original to this repository are the `donutCatalog`, `donutStampsExtra`, `donutStampsIntra`, `outputZernikesRaw` and `outputZernikesAvg`. These are defined in the tasks that create them in `python/lsst/ts/wep/task`.
-  - `--output-run`: Define the name of the new collection that will hold the output from the tasks.
-  - `-p`: Specifies the location of the pipeline configuration file.
-  - `-d`: Define a query to further specify which data in the `collections` we should use.
-    - The query in the example above defines the exposure ID we want to use. Since we have a specific exposure in the gen3 test data that includes the wavefront sensors we specify that number `4021123106000` here in our query.
+
+- `-b`: Specifies the location of the butler repository.
+- `-i`: The `collections` (data) needed for the tasks.
+- `--instrument`: Defines which camera we are using.
+- `--register-dataset-types`: Add the specific datasets types for our tasks to the registry.
+  - Dataset Types that are original to this repository are the `donutCatalog`, `donutStampsExtra`, `donutStampsIntra`, `outputZernikesRaw` and `outputZernikesAvg`. These are defined in the tasks that create them in `python/lsst/ts/wep/task`.
+- `--output-run`: Define the name of the new collection that will hold the output from the tasks.
+- `-p`: Specifies the location of the pipeline configuration file.
+- `-d`: Define a query to further specify which data in the `collections` we should use.
+  - The query in the example above defines the exposure ID we want to use. Since we have a specific exposure in the gen3 test data that includes the wavefront sensors we specify that number `4021123106000` here in our query.
 
 3. If the pipeline ran successfully you can run the following command to see that the name of the new output run is present in the list:
 
@@ -142,7 +144,6 @@ butler query-collections $path_of_ts_wep/tests/testData/gen3TestRepo/
     # |     |     |          |  SW1 |
     # O------------          -------O
 
-
 ## Verify the Calculated Wavefront Error
 
 1. The user can use the `Algorithm.getWavefrontMapEsti()` and `Algorithm.getWavefrontMapResidual()` in `cwfs` module to judge the estimated wavefront error after the solve of transport of intensity equation (TIE). The residual of wavefront map should be low compared with the calculated wavefront map if most of low-order Zernike terms (z4 - z22) have been captured and compensated.
@@ -160,6 +161,25 @@ The rotations for the R04 and R40 wavefront sensors given by the `obs_lsst` pack
 file [`data/lsst/focalPlaneLayout.txt`](https://github.com/lsst-ts/phosim_syseng4/blob/aos/data/lsst/focalplanelayout.txt). This is due to a difference in the rotation direction
 when phosim creates images and the 180 degree difference offsets that and creates the correct orientation with the rest of the focal plane.
 This is shown in the notebook [here](https://github.com/suberlak/AOS/blob/main/AOS_DM-30367_summary.ipynb).
+
+## Note on the orientation of images and masks
+
+***Summary:***
+***Image compensation puts both images into the same orientation.***
+***Pupil masks are generated in the same orientation.***
+***Therefore, we do not need to explicitly rotate images or masks inside the algorithm.***
+
+The Zernike estimation algorithm uses donut images and masks to solve the transport of intensity equation (TIE; see [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X)).
+A pair of intra- and extra-focal donuts is used to approximate the gradient of the donut intensity with respect to the focal direction (i.e. the left-hand side of [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equation 1), while the average of the pair is used to approximate the in-focus image (i.e. $I$ on the right-hand side of [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equation 1).
+
+In order to do this, we must ensure that the donuts are in the same orientation, and apply a mask that corrects for potential vignetting of the intrafocal donut.
+Since the two donuts in the pair are on opposite sides of focus, their orientations are initially rotated by 180 degrees with respect to each other.
+However, before calculating the mean and difference intensities, the donuts are mapped to the pupil plane and "compensated" for the current best-guess of the optical aberrations ([Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equations 9-11).
+This mapping to the pupil plane automatically corrects for the orientation difference.
+Therefore, we do not need to explicitly rotate either image in the algorithm.
+
+In addition, the pupil mask that is applied to both donut images is calculated in the same orientation.
+Therefore, we do not need to rotate the masks either.
 
 ## Build the Document
 
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -5,6 +5,17 @@
 ##################
 Version History
 ##################
+  
+.. _lsst.ts.wep-3.1.0:
+
+-------------
+3.1.0
+-------------
+
+* Added a history to the Algorithm class that stores intermediate products of the algorithm (see `Algorithm.getHistory()`).
+* Fixed the algorithm so that it is once again symmetric with respect to I1 and I2.
+  This involved simplifying the way that mask and image orientation are handled for the extrafocal image (see below).
+* Added the option to create masks in the orientation of the original images by setting `compensated=False` in `CompensableImage.makeMask()`.
 
 .. _lsst.ts.wep-3.0.1:
 
@@ -13,7 +24,7 @@ Version History
 -------------
 
 * Fix ``test_generateDonutCatalogWcsTask.py`` to work with more recent versions of the DM stack.
-
+  
 .. _lsst.ts.wep-3.0.0:
 
 -------------
--- a/python/lsst/ts/wep/cwfs/Algorithm.py
+++ b/python/lsst/ts/wep/cwfs/Algorithm.py
@@ -43,6 +43,7 @@ from lsst.ts.wep.cwfs.Tool import (
     ZernikeAnnularEval,
     ZernikeMaskedFit,
 )
+from lsst.ts.wep.Utility import DefocalType
 from lsst.ts.wep.PlotUtil import plotZernike
 
 
@@ -112,6 +113,9 @@ class Algorithm(object):
         self._dzkdx = None
         self._dzkdy = None
 
+        # Create an attribute to store the algorithm history
+        self._history = dict()
+
     def reset(self):
         """Reset the calculation for the new input images with the same
         algorithm settings."""
@@ -132,6 +136,8 @@ class Algorithm(object):
 
         self.mask_comp_pad = None
         self.mask_pupil_pad = None
+
+        self._history = dict()
 
     def config(self, algoName, inst, debugLevel=0):
         """Configure the algorithm to solve TIE.
@@ -179,6 +185,9 @@ class Algorithm(object):
         self.mask_pupil = None
         self.mask_comp_pad = None
         self.mask_pupil_pad = None
+
+        # Reset the history
+        self._history = dict()
 
     def setDebugLevel(self, debugLevel):
         """Set the debug level.
@@ -493,6 +502,100 @@ class Algorithm(object):
 
         return wfMapWithMask
 
+    def _recordItem(self, item, itemName, outerItr, innerItr=None, debugLevel=0):
+        """Record the item in the algorithm history.
+
+        If you use this method to store new items in the algorithm history, you
+        should update the docstring of the getHistory() method below, which
+        describes the (potential) contents of the algorithm history.
+
+        Parameters
+        ----------
+        item : Any
+            The item to record in the history.
+        itemName : str
+            The name of the item in the history.
+        outerItr : int
+            The iteration of the outer-loop under which to store the item.
+        innerItr : int, optional
+            The iteration of the inner-loop under which to store the item.
+            Inner-loop iterations are stored in another dictionary, which
+            is stored under the key "innerLoop".
+        debugLevel : int, optional
+            The debugLevel at which this item should be recorded.
+            (the default is 0.)
+        """
+        # if debug level too low, do nothing
+        if self.debugLevel < debugLevel:
+            return
+
+        # create the outerItr dictionary if it doesn't exist
+        if outerItr not in self._history:
+            self._history[outerItr] = dict()
+
+        # decide whether to record at the level of the outer- or inner-loop
+        if innerItr is None:
+            self._history[outerItr][itemName] = item
+        else:
+            # create the inner-loop dictionary if it doesn't exist
+            if "innerLoop" not in self._history[outerItr]:
+                self._history[outerItr]["innerLoop"] = dict()
+
+            # create the innerItr dictionary if it doesn't exist
+            if innerItr not in self._history[outerItr]["innerLoop"]:
+                self._history[outerItr]["innerLoop"][innerItr] = dict()
+
+            # record the item
+            self._history[outerItr]["innerLoop"][innerItr][itemName] = item
+
+    def getHistory(self):
+        """Get the algorithm history.
+
+        Returns
+        -------
+        dict
+            Algorithm history.
+
+        Notes
+        -----
+        The algorithm history is a dictionary that contains an entry for each
+        iteration of the outer-loop, beginning with history[0].
+
+        The entry for each outer-loop iteration is itself a dictionary,
+        containing the following keys:
+            - initI1 - the initial I1 image
+            - initI2 - the initial I2 image
+            - compZk - the zernikes used for image compensation (units: nm)
+            - compI1 - the compensated I1 image
+            - compI2 - the compensated I2 image
+            - pupilMask - pupil mask that is applied before zernike calculation
+            - maskedI1 - the masked version of the compensated I1
+            - maskedI2 - the masked version of the compensated I2
+            - residZk - the estimated residual zernikes (units: nm)
+            - residWf - the estimated residual wavefront (units: nm)
+            - totZk - the current best estimate of the zernikes (units: nm)
+            - totWf - the current best estimate of the wavefront (units: nm)
+            - caustic - whether this iteration encountered a caustic
+            - converged - whether the zernikes have converged
+
+        Further, if you are running the FFT algorithm, there is an inner-loop
+        history saved under the key "innerLoop". The inner-loop history is
+        itself a dictionary that contains an entry for every iteration of
+        the inner-loop. The entry for each iteration of the inner-loop is
+        another dictionary, containing the following keys:
+            - initS - the initial "signal"
+            - FFT - the Fast Fourier transform of initS
+            - estWf - the estimated wavefront
+            - estS - the estimated signal
+
+        So for example, if you want to get the estimated signal from the 3rd
+        step of the innerloop during the 2nd step of the outer loop, you
+        would do the following:
+        history = algorithm.getHistory()
+        signal = history[1]["innerLoop"][2]["estS"]
+        """
+        return self._history
+
     def _checkNotItr0(self):
         """Check not in the iteration 0.
 
@@ -547,8 +650,18 @@ class Algorithm(object):
             Optical model. It can be "paraxial", "onAxis", or "offAxis".
         tol : float, optional
             Tolerance of difference of coefficients of Zk polynomials compared
-            with the previous iteration. (the default is 1e-3.)
+            with the previous iteration of outer loop. (the default is 1e-3.)
+
+        Raises
+        ------
+        ValueError
+            Check the that we have an intra- and extra-focal image.
         """
+        if I1.defocalType == I2.defocalType:
+            raise ValueError(
+                f"I1 and I2 are both {I1.defocalType.name}focal. "
+                "Please pass an intra- and extra-focal image."
+            )
 
         # To have the iteration time initiated from global variable is to
         # distinguish the manually and automatically iteration processes.
@@ -657,7 +770,7 @@ class Algorithm(object):
             Optical model. It can be "paraxial", "onAxis", or "offAxis".
         tol : float, optional
             Tolerance of difference of coefficients of Zk polynomials compared
-            with the previous iteration. (the default is 1e-3.)
+            with the previous iteration of outer loop. (the default is 1e-3.)
 
         Returns
         -------
@@ -684,11 +797,12 @@ class Algorithm(object):
                     )
                     sys.exit()
 
-                # Calculate the pupil mask (binary matrix) and related
-                # parameters
+                # Create the pupil masks
+                # These will be applied to compensated images, so we will
+                # create the masks with compensated=True
                 boundaryT = self.getBoundaryThickness()
-                I1.makeMask(self._inst, model, boundaryT, 1)
-                I2.makeMask(self._inst, model, boundaryT, 1)
+                I1.makeMask(self._inst, model, boundaryT, 1, compensated=True)
+                I2.makeMask(self._inst, model, boundaryT, 1, compensated=True)
                 self._makeMasterMask(I1, I2, self.getPoissonSolverName())
 
                 # Load the offAxis correction coefficients
@@ -725,6 +839,10 @@ class Algorithm(object):
             I1.updateImage(I1.getImgInit().copy())
             I2.updateImage(I2.getImgInit().copy())
 
+            # Record the initial images for this iteration
+            self._recordItem(I1.getImg().copy(), "initI1", jj, debugLevel=1)
+            self._recordItem(I2.getImg().copy(), "initI2", jj, debugLevel=1)
+
             if compMode == "zer":
 
                 # Zk coefficient from the previous iteration
@@ -739,23 +857,42 @@ class Algorithm(object):
                 # Add partial feedback of residual estimated wavefront in Zk
                 self.zcomp = self.zcomp + ztmp * feedbackGain
 
+                # Record the zernikes (in nm) used to compensate the images
+                self._recordItem(self.zcomp * 1e9, "compZk", jj, debugLevel=1)
+
                 # Remove the image distortion by forwarding the image to pupil
                 I1.compensate(self._inst, self, self.zcomp, model)
                 I2.compensate(self._inst, self, self.zcomp, model)
 
-            # Check the image condition. If there is the problem, done with
+                # Record the compensated images for this iteration
+                self._recordItem(I1.getImg().copy(), "compI1", jj, debugLevel=1)
+                self._recordItem(I2.getImg().copy(), "compI2", jj, debugLevel=1)
+
+            # Check the image condition. If there is a problem, done with
             # this _singleItr().
             if (I1.isCaustic() is True) or (I2.isCaustic() is True):
                 self.converge[:, jj] = self.converge[:, jj - 1]
                 self.caustic = True
+                self._recordItem(True, "caustic", jj, debugLevel=1)
+                self._recordItem(False, "converged", jj, debugLevel=1)
                 return
 
-            # Correct the defocal images if I1 and I2 are belong to different
-            # sources, which is determined by the (fieldX, field Y)
-            I1, I2 = self._applyI1I2mask_pupil(I1, I2)
+            # Correct the defocal images if I1 and I2 belong to different
+            # sources, which is determined by the (fieldX, fieldY)
+            self._applyI1I2mask_pupil(I1, I2)
+
+            # Record the pupil mask that was used in the previous function,
+            # plus the newly masked images
+            self._recordItem(self.mask_pupil.copy(), "pupilMask", jj, debugLevel=1)
+            self._recordItem(I1.getImg().copy(), "maskedI1", jj, debugLevel=1)
+            self._recordItem(I2.getImg().copy(), "maskedI2", jj, debugLevel=1)
 
             # Solve the Poisson's equation
             self.zc, self.West = self._solvePoissonEq(I1, I2, jj)
+
+            # Record the residual Zernike coefficients (in nm) and wavefront
+            self._recordItem(self.zc.copy() * 1e9, "residZk", jj, debugLevel=1)
+            self._recordItem(self.West.copy(), "residWf", jj, debugLevel=1)
 
             # Record/ calculate the Zk coefficient and wavefront
             if compMode == "zer":
@@ -768,6 +905,12 @@ class Algorithm(object):
                     yoSensor,
                     self.getObsOfZernikes(),
                 )
+
+                # Record the total Zernike coefficients (in nm) and wavefront
+                self._recordItem(
+                    self.converge[:, jj].copy() * 1e9, "totZk", jj, debugLevel=1
+                )
+                self._recordItem(self.wcomp.copy(), "totWf", jj, debugLevel=1)
 
         else:
             # Once we run into caustic, stop here, results may be close to real
@@ -792,10 +935,15 @@ class Algorithm(object):
             if diffZk < tol:
                 stopItr = True
 
+        # Record the convergence
+        if not self.caustic:
+            self._recordItem(False, "caustic", jj, debugLevel=1)
+            self._recordItem(stopItr, "converged", jj, debugLevel=1)
+
         # Update the current iteration time
         self.currentItr += 1
 
-        # Show the Zk coefficients in interger in each iteration
+        # Show the Zk coefficients in integer in each iteration
         if self.debugLevel >= 2:
             print("itr = %d, z4-z%d" % (jj, self.getNumOfZernikes()))
             print(np.rint(self.zer4UpNm))
@@ -901,8 +1049,14 @@ class Algorithm(object):
             S = Sini.copy()
             for jj in range(self.getNumOfInnerItr()):
 
-                # Calculate FT{S}
+                # Record the initial image
+                self._recordItem(S.copy(), "initS", iOutItr, jj, debugLevel=1)
+
+                # Calculate FFT{S}
                 SFFT = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(S)))
+
+                # Record FFT{S}
+                self._recordItem(SFFT.copy(), "FFT", iOutItr, jj, debugLevel=1)
 
                 # Calculate W by W=IFT{ FT{S}/(-4*pi^2*(u^2+v^2)) }
                 W = np.fft.fftshift(
@@ -920,6 +1074,9 @@ class Algorithm(object):
                 West = West - offset
                 West[self.mask_pupil == 0] = 0
 
+                # Record the estimated wavefront
+                self._recordItem(West.copy(), "estWf", iOutItr, jj, debugLevel=1)
+
                 # Set dWestimate/dn = 0 around boundary
                 WestdWdn0 = West.copy()
 
@@ -932,8 +1089,10 @@ class Algorithm(object):
 
                 kernel = np.ones((1 + 2 * boundaryT, 1 + 2 * boundaryT))
                 tmp = convolve2d(West * ApringIn, kernel, mode="same")
-                tmp /= convolve2d(ApringIn, kernel, mode="same")
-                WestdWdn0[borderx, bordery] = tmp[borderx, bordery]
+                tmp_norm = convolve2d(ApringIn, kernel, mode="same")
+                WestdWdn0[borderx, bordery] = (
+                    tmp[borderx, bordery] / tmp_norm[borderx, bordery]
+                )
 
                 # Take Laplacian to find sensor signal estimate (Delta W = S)
                 del2W = laplace(WestdWdn0) / dOmega
@@ -941,6 +1100,9 @@ class Algorithm(object):
                 # Extend the dimension of signal to the order of 2 for "fft" to
                 # use
                 Sest = padArray(del2W, padDim)
+
+                # Record the estimated signal
+                self._recordItem(Sest.copy(), "estS", iOutItr, jj, debugLevel=1)
 
                 # Put signal back inside boundary, leaving the rest of
                 # Sestimate
@@ -963,8 +1125,8 @@ class Algorithm(object):
             # Use the integration method by serial expansion to solve the
             # Poisson's equation
 
-            # Calculate I0 and dI
-            I0, dI = self._getdIandI(I1, I2)
+            # Calculate dI and I0
+            dI, I0 = self._getdIandI0(I1, I2)
 
             # Get the x, y coordinate in mask. The element outside mask is 0.
             xSensor, ySensor = self._inst.getSensorCoor()
@@ -1038,13 +1200,14 @@ class Algorithm(object):
             Approximated wavefront signal.
         """
 
-        # Check the condition of images
-        I1image, I2image = self._checkImageDim(I1, I2)
+        # Calculate I0 and dI
+        dI, I0 = self._getdIandI0(I1, I2)
 
         # Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be
-        # -(1/delta z)*(I1-I2)/(I1+I2)
-        num = I1image - I2image
-        den = I1image + I2image
+        # -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)
+        # we will ignore the (1/delta z) until later in this method
+        num = -dI
+        den = 2 * I0
 
         # Define the effective minimum central signal element by the threshold
         # ( I0=(I1+I2)/2 )
@@ -1081,13 +1244,12 @@ class Algorithm(object):
 
         return Sout
 
-    def _getdIandI(self, I1, I2):
-        """Calculate the central image and differential image to be used in the
-        serial expansion method.
+    def _getdIandI0(self, I1, I2):
+        """Calculate the differential image and central image.
 
-        It is noted that the images are assumed to be co-center already. And
-        the intra-/ extra-focal image can overlap with one another after the
-        rotation of 180 degree.
+        It is noted that the images are assumed to be co-center already, and
+        that the images have already been compensated, so that the orientation
+        of the extrafocal matches that of the intrafocal image.
 
         Parameters
         ----------
@@ -1099,25 +1261,27 @@ class Algorithm(object):
         Returns
         -------
         numpy.ndarray
-            Image data of I0.
+            Differential image, dI = Extra - Intra
         numpy.ndarray
-            Differential image (dI) of I0.
+            I0 = (Extra + Intra) / 2
         """
 
-        # Check the condition of images
-        I1image, I2image = self._checkImageDim(I1, I2)
+        # Check the image dimensions
+        self._checkImageDim(I1, I2)
 
         # Calculate the central image and differential image
-        I0 = (I1image + I2image) / 2
-        dI = I2image - I1image
+        dI = I1.getImg() - I2.getImg()
+        I0 = (I1.getImg() + I2.getImg()) / 2
 
-        return I0, dI
+        # if I2 is extrafocal, flip the sign of dI
+        if I2.defocalType == DefocalType.Extra:
+            dI = -dI
+
+        return dI, I0
 
     def _checkImageDim(self, I1, I2):
         """Check the dimension of images.
 
-        It is noted that the I2 image is rotated by 180 degree.
-
         Parameters
         ----------
         I1 : CompensableImage
@@ -1125,19 +1289,11 @@ class Algorithm(object):
         I2 : CompensableImage
             Intra- or extra-focal image.
 
-        Returns
-        -------
-        numpy.ndarray
-            I1 defocal image.
-        numpy.ndarray
-            I2 defocal image. It is noted that the I2 image is rotated by 180
-            degree.
-
         Raises
         ------
-        Exception
+        ValueError
             Check the dimension of images is n by n or not.
-        Exception
+        ValueError
             Check two defocal images have the same size or not.
         """
 
@@ -1146,19 +1302,10 @@ class Algorithm(object):
         m2, n2 = I2.getImg().shape
 
         if m1 != n1 or m2 != n2:
-            raise Exception("Image is not square.")
+            raise ValueError("Image is not square.")
 
         if m1 != m2 or n1 != n2:
-            raise Exception("Images do not have the same size.")
-
-        # Define I1
-        I1image = I1.getImg()
-
-        # Rotate the image by 180 degree through rotating two times of 90
-        # degree
-        I2image = np.rot90(I2.getImg(), k=2)
-
-        return I1image, I2image
+            raise ValueError("Images do not have the same size.")
 
     def _makeMasterMask(self, I1, I2, poissonSolver=None):
         """Calculate the common mask of defocal images.
@@ -1188,8 +1335,9 @@ class Algorithm(object):
             self.mask_pupil_pad = padArray(self.mask_pupil, padDim)
 
     def _applyI1I2mask_pupil(self, I1, I2):
-        """Correct the defocal images if I1 and I2 are belong to different
-        sources.
+        """Mask the images if I1 and I2 belong to different sources.
+
+        Note I1 and I2 are mutated in-place.
 
         (There is a problem for this actually. If I1 and I2 come from different
         sources, what should the correction of TIE be? At this moment, the
@@ -1202,13 +1350,6 @@ class Algorithm(object):
             Intra- or extra-focal image.
         I2 : CompensableImage
             Intra- or extra-focal image.
-
-        Returns
-        -------
-        numpy.ndarray
-            Corrected I1 image.
-        numpy.ndarray
-            Corrected I2 image.
         """
 
         # Get the overlap region of images and do the normalization.
@@ -1216,19 +1357,11 @@ class Algorithm(object):
 
             # Get the overlap region of image
             I1.updateImage(I1.getImg() * self.mask_pupil)
-
-            # Rotate mask_pupil by 180 degree through rotating two times of 90
-            # degree because I2 has been rotated by 180 degree already.
-            I2.updateImage(I2.getImg() * np.rot90(self.mask_pupil, 2))
+            I2.updateImage(I2.getImg() * self.mask_pupil)
 
             # Do the normalization of image.
             I1.updateImage(I1.getImg() / np.sum(I1.getImg()))
             I2.updateImage(I2.getImg() / np.sum(I2.getImg()))
-
-        # Return the correct images. It is noted that there is no need of
-        # vignetting correction.
-        # This is after masking already in _singleItr() or itr0().
-        return I1, I2
 
     def _reset(self, I1, I2):
         """Reset the iteration time of outer loop and defocal images.
@@ -1269,7 +1402,7 @@ class Algorithm(object):
 
     def outZer4Up(self, unit="nm", filename=None, showPlot=False):
         """Put the coefficients of normal/ annular Zernike polynomials on
-        terminal or file ande show the image if it is needed.
+        terminal or file and show the image if it is needed.
 
         Parameters
         ----------
--- a/python/lsst/ts/wep/cwfs/CompensableImage.py
+++ b/python/lsst/ts/wep/cwfs/CompensableImage.py
@@ -462,9 +462,6 @@ class CompensableImage(object):
         # I(x, y)/I'(x', y') = J = (dx'/dx)*(dy'/dy) - (dx'/dy)*(dy'/dx)
         self.updateImage(lutIp * J)
 
-        if self.defocalType == DefocalType.Extra:
-            self.updateImage(np.rot90(self.getImg(), k=2))
-
         # Put NaN to be 0
         imgCompensate = self.getImg()
         imgCompensate[np.isnan(imgCompensate)] = 0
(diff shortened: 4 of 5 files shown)


tests/cwfs/test_algorithm.py
```diff
@@ -312,6 +312,25 @@ def testRunItOfFft(self):
...
 
+    def testAlgLog(self):
+
+        self.algoExp.reset()
+        self.algoExp.setDebugLevel(0)
+        self.assertIsNone(self.algoExp.getLog())
+
+        self.algoExp.setDebugLevel(3)
+        self.algoExp.runIt(self.I1, self.I2, self.opticalModel, tol=1e-3)
+        self.assertIsTrue(isinstance(self.algoExp.getLog(), dict))
```
suberlak: Apart from checking that it's a dict do you want to check that it has the keys that you expect (i..e that it got populated when debugLevel is 3) ?

tests/cwfs/test_algorithm.py
```diff
@@ -20,13 +20,13 @@
...
 import unittest
 
-from lsst.ts.wep.cwfs.Instrument import Instrument
-from lsst.ts.wep.cwfs.CompensableImage import CompensableImage
+import numpy as np
 from lsst.ts.wep.cwfs.Algorithm import Algorithm
-from lsst.ts.wep.Utility import getModulePath, getConfigDir, DefocalType, CamType
+from lsst.ts.wep.cwfs.CompensableImage import CompensableImage
+from lsst.ts.wep.cwfs.Instrument import Instrument
+from lsst.ts.wep.Utility import CamType, DefocalType, getConfigDir, getModulePath
```
suberlak: I wonder if this is a result of running black? If so, it would be better for those changes to be a separate PR not to be confusing (because there isn't actual content changed here, it just seems rearranged, right?)

suberlak: If these changes are not necessitated to pass flake and CI then I would undo

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -1494,3 +1493,7 @@ def makeMask(self, inst, model, boundaryT, maskScalingFactorLocal):
 
             # padded mask for use at the offset planes
             self.mask_comp = self.mask_comp * mask_comp_ii
+
+        if not compensated and self.defocalType == DefocalType.Extra:
```
suberlak, line 1505: According to docstring above, it would seem that  this line should be `if compensated and ...` . Or should the docstring say "If compensated is False"?

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -1434,6 +1428,11 @@ def makeMask(self, inst, model, boundaryT, maskScalingFactorLocal):
             zero.
         maskScalingFactorLocal : float
             Mask scaling factor (for fast beam) for local correction.
+        compensated: bool, optional
+            Whether to calculate masks for the original or compensated image.
+            If compensated is True, then extrafocal masks will be rotated by
```
suberlak: See the comment below (should it say `If compensated is False` ?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1040,13 +1107,14 @@ def _createSignal(self, I1, I2, cliplevel):
...
+        # Calculate I0 and dI
+        dI, I0 = self._getdIandI0(I1, I2)
 
         # Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be
-        # -(1/delta z)*(I1-I2)/(I1+I2)
-        num = I1image - I2image
-        den = I1image + I2image
+        # -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)
+        # we will ignore the (1/delta z) until later in this method
+        num = -dI
```
suberlak, line 1209: just making sure  - does it pass the 'fft' unit tests?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -969,7 +1036,7 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
             # Poisson's equation
 
             # Calculate I0 and dI
-            I0, dI = self._getdIandI(I1, I2)
+            dI, I0 = self._getdIandI0(I1, I2)
```
suberlak, line 1129: That's a good idea to align the returned elements with the function name!

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -23,9 +23,19 @@
...
+from lsst.ts.wep.cwfs.Instrument import Instrument
+from lsst.ts.wep.cwfs.Tool import (
+    ZernikeAnnularEval,
+    ZernikeMaskedFit,
+    extractArray,
+    padArray,
+)
+from lsst.ts.wep.ParamReader import ParamReader
+from lsst.ts.wep.PlotUtil import plotZernike
+from lsst.ts.wep.Utility import DefocalType
```
suberlak: This doesn't seem to make any changes. If it's a result of black, see comment below - should be in a separate PR. Probably not needed to pass CI given that it passed before (?)

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -744,24 +767,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
             if (I1.isCaustic() is True) or (I2.isCaustic() is True):
                 self.converge[:, jj] = self.converge[:, jj - 1]
                 self.caustic = True
+                self._logItem(True, "caustic", jj, debugLevel=1)
+                self._logItem(False, "converged", jj, debugLevel=1)
                 return
 
-            # Correct the defocal images if I1 and I2 are belong to different
-            # sources, which is determined by the (fieldX, field Y)
-            I1, I2 = self._applyI1I2mask_pupil(I1, I2)
```
suberlak, line 755: Was this not necessary because the internal state of compensable images gets updated within this function?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1099,40 +1166,34 @@ def _getdIandI(self, I1, I2):
...
-            Image data of I0.
+            Differential image, dI = Extra - Intra
         numpy.ndarray
-            Differential image (dI) of I0.
+            I0 = (Extra + Intra) / 2
         """
 
-        # Check the condition of images
-        I1image, I2image = self._checkImageDim(I1, I2)
+        # Check the image dimensions
```
suberlak, line 1269: Thanks, I didn't think it was a good idea to have a "checking" function return anything besides an error if the check doesn't go well...

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -1494,3 +1493,7 @@ def makeMask(self, inst, model, boundaryT, maskScalingFactorLocal):
 
             # padded mask for use at the offset planes
             self.mask_comp = self.mask_comp * mask_comp_ii
+
+        if not compensated and self.defocalType == DefocalType.Extra:
+            self.mask_pupil = np.rot90(self.mask_pupil, 2)
+            self.mask_comp = np.rot90(self.mask_comp, 2)
```
suberlak, line 1507: So this is done here instead of at the `_applyI1I2mask_pupil` stage (from where it was removed)? Perhaps worth mentioning that as a comment

tests/cwfs/test_algorithm.py
suberlak: ... and that the content of keys is what you'd assume (eg. `residZk`  is an array with 19 values, etc..)

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -23,24 +23,19 @@
...
+import numpy as np
+from galsim.utilities import horner2d
+from lsst.ts.wep.cwfs.Image import Image
+from lsst.ts.wep.cwfs.Tool import (ZernikeAnnularGrad, ZernikeAnnularJacobian,
+                                   extractArray, padArray)
+from lsst.ts.wep.ParamReader import ParamReader
+from lsst.ts.wep.Utility import CentroidFindType, DefocalType
 from scipy.interpolate import RectBivariateSpline
+from scipy.ndimage import (binary_dilation, binary_erosion,
+                           generate_binary_structure, iterate_structure)
```
suberlak: see comment above -  if these are purely black changes, it should be a separate PR (since there aren't any content changes that I can see, merely a rearrangement)

python/lsst/ts/wep/cwfs/CompensableImage.py
suberlak: Or is it instead of `_checkImageDim` rotation?

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -462,9 +457,6 @@ def compensate(self, inst, algo, zcCol, model):
         # I(x, y)/I'(x', y') = J = (dx'/dx)*(dy'/dy) - (dx'/dy)*(dy'/dx)
         self.updateImage(lutIp * J)
 
-        if self.defocalType == DefocalType.Extra:
```
suberlak, line 465: Is this step done now somewhere else? Why is it not needed ?

suberlak: Would be useful to have some sort of a flowchart diagram saying what's the rotation of I1,I2 as we go through the algorithm (what was it before, and what your changes make), like https://confluence.lsstcorp.org/display/LTS/Algorithm+Flow+Diagram  (just because it's always been quite confusing to have so many rotations...)

suberlak (changes requested):
Since it's a major change to one of the core elements of `ts_wep`, it would be good to get feedback from @teweitsai

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1337,3 +1373,85 @@ def outZer4Up(self, unit="nm", filename=None, showPlot=False):
         if showPlot:
             zkIdx = range(4, len(z) + 4)
             plotZernike(zkIdx, z, unit)
+
```
jbkalmbach: Maybe move these functions up near the top of the class since they are used throughout the rest of the class.

python/lsst/ts/wep/cwfs/CompensableImage.py
jbkalmbach: Also note in the description what the default option is. See docstring for `_showProjection` above.

jbkalmbach (commented):
I don't have too much more to add beyond Chris' comments but just left a couple small notes. Looks pretty good to me.

doc/versionHistory.rst
```diff
@@ -6,6 +6,18 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-2.7.0:
+
+-------------
+2.7.0
+-------------
+
+* Added internal logging to the Algorithm class (see `Algorithm.getLog`).
```
teweitsai: `Algorithm.getLog` --> `Algorithm.getLog()`.

doc/versionHistory.rst
```diff
@@ -6,6 +6,18 @@
...
+* Fixed the algorithm so that it is once again symmetric with respect to I1 and I2.
+  This involved simplifying the way that mask and image orientation are handled for the extrafocal image (see below).
+* The extrafocal mask now matches the orientation of the extrafocal image.
+  If you want the mask to match the orientation of the image post-compensation (i.e. after the 180 deg rotation), you can set `compensated=True` in `CompensableImage.makeMask`.
```
teweitsai: `CompensableImage.makeMask()`?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -181,6 +186,9 @@ def config(self, algoName, inst, debugLevel=0):
         self.mask_comp_pad = None
         self.mask_pupil_pad = None
 
+        # reset the log
```
teweitsai: reset --> Reset

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -552,7 +560,17 @@ def runIt(self, I1, I2, model, tol=1e-3):
...
+        Raises
+        ------
+        Exception
+            Check the that we have an intra- and extra-focal image.
         """
+        if I1.defocalType == I2.defocalType:
+            raise Exception(
+                f"I1 and I2 are both {I1.defocalType.name}focal. "
+                "Please pass an intra- and extra-focal image."
+            )
```
teweitsai: Please use the more detailed exception type instead such as `ValueError`:
https://docs.python.org/3/library/exceptions.html

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -688,11 +706,12 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
                     )
                     sys.exit()
 
-                # Calculate the pupil mask (binary matrix) and related
-                # parameters
+                # create the masks
+                # these will be applied to compensated images, so we will create
+                # the masks with compensated=True
```
teweitsai:
```
# Create the masks
# These will be applied to compensated images, so we will create
# the masks with compensated=True
```

teweitsai: Why do you want to remove the word of `pupil`? In WEP, you have several coordinate systems need to consider.

python/lsst/ts/wep/cwfs/CompensableImage.py
```diff
@@ -1434,6 +1428,11 @@ def makeMask(self, inst, model, boundaryT, maskScalingFactorLocal):
             zero.
         maskScalingFactorLocal : float
             Mask scaling factor (for fast beam) for local correction.
+        compensated: bool, optional
+            Whether to calculate masks for the original or compensated image.
+            If compensated is True, then extrafocal masks will be rotated by
+            180 deg, to match the rotation that occurs to the extrafocal image
+            during compensation.
```
teweitsai: You need to document the default value in doc string.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -730,6 +749,10 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
             I1.updateImage(I1.getImgInit().copy())
             I2.updateImage(I2.getImgInit().copy())
 
+            # log the initial images for this iteration
```
teweitsai:
```
# Log the initial images for this iteration
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -744,24 +767,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
                 # Add partial feedback of residual estimated wavefront in Zk
                 self.zcomp = self.zcomp + ztmp * feedbackGain
 
+                # log the zernikes (in nm) used to compensate the images
```
teweitsai:
```
# Log the Zernikes (in nm) used to compensate the images
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -744,24 +767,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
+                self._logItem(self.zcomp * 1e9, "compZk", jj, debugLevel=1)
+
                 # Remove the image distortion by forwarding the image to pupil
                 I1.compensate(self._inst, self, self.zcomp, model)
                 I2.compensate(self._inst, self, self.zcomp, model)
 
-            # Check the image condition. If there is the problem, done with
+                # log the compensated images for this iteration
```
teweitsai:
```
# Log the compensated images for this iteration
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -744,24 +767,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
                 return
 
-            # Correct the defocal images if I1 and I2 are belong to different
-            # sources, which is determined by the (fieldX, field Y)
-            I1, I2 = self._applyI1I2mask_pupil(I1, I2)
+            # Correct the defocal images if I1 and I2 belong to different
+            # sources, which is determined by the (fieldX, fieldY)
+            self._applyI1I2mask_pupil(I1, I2)
+
+            # log the pupil mask that was used in the previous function, plus
```
teweitsai:
```
# Log the pupil mask that was used in the previous function, plus
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -744,24 +767,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
+            # the newly masked images
+            self._logItem(self.mask_pupil.copy(), "pupilMask", jj, debugLevel=1)
+            self._logItem(I1.getImg().copy(), "maskedI1", jj, debugLevel=1)
+            self._logItem(I2.getImg().copy(), "maskedI2", jj, debugLevel=1)
 
             # Solve the Poisson's equation
             self.zc, self.West = self._solvePoissonEq(I1, I2, jj)
 
+            # log the residual zernikes (in nm) and wavefront
```
teweitsai:
```
# Log the residual zernikes (in nm) and wavefront
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -774,6 +816,12 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
                     self.getObsOfZernikes(),
                 )
 
+                # log the total zernikes (in nm) and wavefront
```
teweitsai:
```
# Log the total zernikes (in nm) and wavefront
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -797,10 +845,15 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
             if diffZk < tol:
                 stopItr = True
 
+        # log the convergence
```
teweitsai:
```
# Log the convergence
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -906,9 +959,15 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
             S = Sini.copy()
             for jj in range(self.getNumOfInnerItr()):
 
+                # log the initial image
```
teweitsai:
```
# Log the initial image
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -906,9 +959,15 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
...
+                self._logItem(S.copy(), "initS", iOutItr, jj, debugLevel=1)
+
                 # Calculate FT{S}
```
teweitsai: FFT

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -906,9 +959,15 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
...
                 SFFT = np.fft.fftshift(np.fft.fft2(np.fft.fftshift(S)))
 
+                # log FT{S}
```
teweitsai:
```
# Log FFT{S}
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -925,6 +984,9 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
                 West = West - offset
                 West[self.mask_pupil == 0] = 0
 
+                # log the estimated wavefront
```
teweitsai:
```
# Log the estimated wavefront
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -947,6 +1011,9 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
                 # use
                 Sest = padArray(del2W, padDim)
 
+                # log the estimated signal
```
teweitsai:
```
# Log the estimated signal
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1151,15 +1212,6 @@ def _checkImageDim(self, I1, I2):
         if m1 != m2 or n1 != n2:
             raise Exception("Images do not have the same size.")
```
teweitsai: This is a good time to use the `ValueError` here instead.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1337,3 +1373,85 @@ def outZer4Up(self, unit="nm", filename=None, showPlot=False):
...
+            The iteration of the outer-loop under which to store the item.
+        innerItr : int, optional
+            The iteration of the inner-loop under which to store the item.
+        debugLevel : int, optional
+            The debugLevel at which this item should be logged.
+        """
+        if self.debugLevel >= debugLevel:
+            # create the log if it doesn't exist
+            if self.log is None:
+                self.log = {}
```
teweitsai: self.log = dict()

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1337,3 +1373,85 @@ def outZer4Up(self, unit="nm", filename=None, showPlot=False):
...
+                # create the inner-loop dictionary if it doesn't exist
+                if "innerLoop" not in self.log[outerItr]:
+                    self.log[outerItr]["innerLoop"] = {}
+
+                # create the innerItr dictionary if it doesn't exist
+                if innerItr not in self.log[outerItr]["innerLoop"]:
+                    self.log[outerItr]["innerLoop"][innerItr] = {}
+
+                # log the item
+                self.log[outerItr]["innerLoop"][innerItr][itemName] = item
```
teweitsai: This `_logItem()` has 3 indentation layers here. Please consider to refactor it to decrease the number of layers to improve the maintainability.

teweitsai: BTW, please use the `dict()` instead of `{}` to assign the default value to improve the readability.

teweitsai (commented):
This PR looks good to me. I think:

1. This PR contains the new feature of `log` and code refactoring. However, each PR should be for a single target to make the reviewer knows what to review (and make his/her live easier). Under this condition, you could use two or more PRs instead. If you think that is not necessary, another option is to rewrite the `git` history to have several comments and each commit is for different purpose for the reviewer to understand the details. BTW, if the code change is large, squash all changes to a single commit is not a good option in PR.

2. The name of `self.log` is a bad idea because in T&S control software, we use the `log` for the `logging` instance instead of `dict`. You could see an example here:
https://github.com/lsst-ts/ts_mtaos/blob/develop/python/lsst/ts/mtaos/model.py#L181-L184

You could see the base of control system (`ts_salobj`) as well:
https://github.com/lsst-ts/ts_salobj/blob/develop/python/lsst/ts/salobj/controller.py#L221

Therefore, please rename the `self.log` to something else to avoid the confusion with other control systems.

3. The sentence of comment is better to begin from `capital`: `# This is a comment` instead of `# this is a comment`.

4. I noticed that the `rot90()` were removed at multiple places. Is the calculation of wavefront error still right?

5. Please make sure all the Jenkins tests pass. Do not delete the failed tests unless you know what you are doing.

Thanks!

teweitsai:
I just noticed there is no unit tests for the new added functions in this PR. For each new function, please add the related unit test as well. Thanks!

On the Jenkins, there is the coverage report. You could use that function to see how many lines of new code had been tested:
https://tssw-ci.lsst.org/job/LSST_Telescope-and-Site/job/ts_wep/job/PR-152/1/Coverage_20Report/

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: Yeah I agree the way I wrote this sounds a little confusing. It is because the masks are calculated in the compensated orientation. I will clear this up.

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: Added the default and clarified the explanation.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: Yeah there were several rotations I was able to eliminate.

There are essentially two orientations:
1. the _Original Orientation_, where the extrafocal image is rotated 180 from the intrafocal image
2. the _Compensated Orientation_, where the extrafocal image lines up with the intrafocal image

How it works now:
- the masks are generated in the _Compensated Orientation_, and therefore never need to be rotated in the algorithm
- the extrafocal image is *automatically* rotated to the _Compensated Orientation_ *during compensation*. 
- everything happens in the _Compensated Orientation_, so you never need to explicitly rotate anything

Basically, the idea is put everything in the _Compensated Orientation_ at the beginning, and remove all of the other extraneous rotations!

---
Below I explain what it was doing before. It's not necessary, but just in case you want to know.

Image compensation already rotates the extrafocal image so that its orientation matches the intrafocal image. The extra rotation that I deleted above rotated the extrafocal image *back* to the _Original Orientation_. This had 2 downstream effects in the algorithm:

1. in `_applyI1I2mask_pupil`, it had to rotate the pupil mask to the _Original Orientation_ to match the extrafocal image which was now back in the _Original Orientation_. However, when doing this, it assumed that I2 was the extrafocal image.

2. When calculating the signal and difference image, it needed to rotate the extrafocal image back to the _Compensated Orientation_ to match the intrafocal image. It did this in `_checkImageDim`, which was really confusing! It again assumes that I2 is the extrafocal image!

So by removing this extra rotation in `CompensableImage.compensate()`, we can just keep everything in the _Compensated Orientation_ and remove these latter two rotations, which simplifies everything a lot. It also allowed me to remove two places that were assuming I2 was the extrafocal image (but those weren't the only places this assumption was made!)

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Correct. Explicitly returning the images is unnecessary, and is misleading, since it suggests that the original objects are unchanged

tests/cwfs/test_algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: It's actually a little different. This rotation is never actually relevant to the algorithm, since the algorithm calls makeMask with `compensated=True`. It just ensures that if you are looking at the original images, their masks have the correct orientation.

doc/versionHistory.rst
jfcrenshaw: Done

doc/versionHistory.rst
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Ah I was just trying to match the pattern in the rest of the file. I will promote this to a ValueError!

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Because it creates `mask_pupil` _and_ `mask_comp`, so I thought it was misleading. But I can add the word pupil back in if you think it is appropriate?

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: I dont think I understand this comment? You want me to say this is logging the FFT? Isn't this logging the image before the FFT?

jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
teweitsai: Please help to add the word of `pupil` back for the clarification. Thanks!

tests/cwfs/test_algorithm.py
jfcrenshaw: Added checks for all the keys present

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Yep! I made sure every change left the results unchanged

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done!

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done!

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

jfcrenshaw:
Made lots of small changes in response to the above comments. The only significant change was renaming Log -> History to not cause confusion with other logging features. 

The only failing unit test has to do with donut centroids loaded from the butler. It's not clear to me how any of my changes touch that at all. Looking into this with Chris.

jfcrenshaw:
Fixed the failing test by defaulting the masks to the compensated orientation again. I thought this might happen. You can still request the masks in the orientation of the original image by setting `compensated=False` in `makeMask`.

jfcrenshaw:
I believe I have addressed all of the comments above, and all tests now pass.

I am requesting a review again. There is one potential issue: whether I need to split this into two PRs. I developed these two changes (the algorithm history, and restoring the I1/I2 symmetry) in parallel, because both of these were changing in response to the other. Now that this has converged to a final state, I could spend some time pulling the changes out into two separate PRs if needed, but would prefer not to spend this time unless we really need to. But if it's necessary, let me know!

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -112,6 +113,9 @@ def __init__(self, algoDir):
         self._dzkdx = None
         self._dzkdy = None
 
+        # Create an attribute to store the algorithm history
+        self.history = None
```
teweitsai: You can set this as a dict first: `self.history = dict()` to let the developer notices this will be an instance of `dict` in a latter time. Then, do the similar thing at the `self.reset()` and `self.config()`. After this, you could remove the lines 525-527 at `self._recordItem()` for:

```
        # create the history if it doesn't exist
         if self.history is None:
             self.history = dict()
```

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+        item : Any
+            The item to record in the history.
+        itemName : str
+            The name of the item in the history.
+        outerItr : int
+            The iteration of the outer-loop under which to store the item.
+        innerItr : int, optional
+            The iteration of the inner-loop under which to store the item.
+        debugLevel : int, optional
+            The debugLevel at which this item should be saved.
```
teweitsai: For the optional item, you need to put the default value in the doc string such as "(the default is xxx)".

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+        """
+        # if debug level too low, do nothing
+        if self.debugLevel < debugLevel:
+            return
+
+        # create the history if it doesn't exist
+        if self.history is None:
+            self.history = dict()
```
teweitsai: You may not need this. See above comment.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+        if outerItr not in self.history:
+            self.history[outerItr] = dict()
+
+        # decide whether to record at the level of the outer- or inner-loop
+        if innerItr is None:
+            self.history[outerItr][itemName] = item
+        else:
+            # create the inner-loop dictionary if it doesn't exist
+            if "innerLoop" not in self.history[outerItr]:
+                self.history[outerItr]["innerLoop"] = dict()
```
teweitsai: You need to mention this hard-coded "innerLoop" in the doc string of function. And the type of value is `dict`.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
 
         return wfMapWithMask
 
+    def _recordItem(self, item, itemName, outerItr, innerItr=None, debugLevel=0):
```
teweitsai, line 505: You can refer to the `self.getHistory()` in doc string for the possible keys.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+            # create the innerItr dictionary if it doesn't exist
+            if innerItr not in self.history[outerItr]["innerLoop"]:
+                self.history[outerItr]["innerLoop"][innerItr] = dict()
+
+            # record the item
+            self.history[outerItr]["innerLoop"][innerItr][itemName] = item
+
+    def getHistory(self):
+        """Get the algorithm history.
+
```
teweitsai, line 553: I am not sure @jbkalmbach wants to have this `get` function or not since he is refactoring the `ts_wep` to remove the `get`/`set` functions. In pythonic style, people do not intend to use the `get`/`set` function. I am more JAVA style when I developed this package and used to the `get`/`set` functions. You guys could discuss what to do here.

If you want to have the `get` method, you could rename the `self.history` to `self._history` to imply this is for the internal use. If you want to remove the `get` method, you can put the doc string in this function to other place because I do not want to lose it, which is very useful.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,97 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+            - compI2 - the compensated I2 image
+            - pupilMask - pupil mask that is applied before zernike calculation
+            - maskedI1 - the masked version of the compensated I1
+            - maskedI2 - the masked version of the compensated I2
+            - residZk - the estimated residual zernikes
+            - residWf - the estimated residual wavefront
+            - totZk - the current best estimate of the zernikes
+            - totWf - the current best estimate of the wavefront
+            - caustic - whether this iteration encountered a caustic
+            - converged - whether the zernikes have converged
```
teweitsai: Based on this, the `itemName` in `self._recordItem()` is restricted. It is not a random value. Therefore, you may need to add the appropriate doc string in `self._recordItem()`. Another method is to do the programmatic check before you want to record anything such as to check the available `itemName` and raise the error or something similar.

teweitsai: For the physical value such as `zk` and others, put the unit in doc string.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -739,24 +854,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
+                self._recordItem(True, "caustic", jj, debugLevel=1)
+                self._recordItem(False, "converged", jj, debugLevel=1)
                 return
 
-            # Correct the defocal images if I1 and I2 are belong to different
-            # sources, which is determined by the (fieldX, field Y)
-            I1, I2 = self._applyI1I2mask_pupil(I1, I2)
+            # Correct the defocal images if I1 and I2 belong to different
+            # sources, which is determined by the (fieldX, fieldY)
+            self._applyI1I2mask_pupil(I1, I2)
```
teweitsai, line 882: I think in the pythonic style, if the objects `I1` and `I2` will be modified in the function, we prefer to expose this information to be:

```
I1, I2 = self._applyI1I2mask_pupil(I1, I2)
```

But in the C/C++, we do not do this.

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1038,13 +1197,14 @@ def _createSignal(self, I1, I2, cliplevel):
...
-        I1image, I2image = self._checkImageDim(I1, I2)
+        # Calculate I0 and dI
+        dI, I0 = self._getdIandI0(I1, I2)
 
         # Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be
-        # -(1/delta z)*(I1-I2)/(I1+I2)
-        num = I1image - I2image
-        den = I1image + I2image
+        # -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)
+        # we will ignore the (1/delta z) until later in this method
```
teweitsai, line 1208: Do you have a good reason for this? I felt the original implementation is more straightforward for the math. Maybe it looks slower, but it is more straightforward to understand the math behind (for the future's developers/maintainers).

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -1202,34 +1345,19 @@ def _applyI1I2mask_pupil(self, I1, I2):
...
+            I2.updateImage(I2.getImg() * self.mask_pupil)
 
             # Do the normalization of image.
             I1.updateImage(I1.getImg() / np.sum(I1.getImg()))
             I2.updateImage(I2.getImg() / np.sum(I2.getImg()))
 
-        # Return the correct images. It is noted that there is no need of
-        # vignetting correction.
-        # This is after masking already in _singleItr() or itr0().
-        return I1, I2
```
teweitsai, line 1231: I prefer to return the `I1` and `I2` if I am working in Python. For C/C++, I will not return them.

If you do not want to return them, you can mention the `I1` and `I2` will be modified in the doc string.

tests/cwfs/test_algorithm.py
```diff
@@ -314,6 +314,90 @@ def testRunItOfFft(self):
         zk = self.algoFft.getZer4UpInNm()
         self.assertEqual(int(zk[7]), -192)
 
+    def testSymmetryI1I2(self):
+        # Test that the algorithm gives same results when you swap I1 <-> I2.
+        pass
+
```
teweitsai: You forgot to implement this `testSymmetryI1I2()`.

tests/cwfs/test_algorithm.py
```diff
@@ -314,6 +314,90 @@ def testRunItOfFft(self):
...
+    def testAlgHistory(self):
+
+        # we will run all history tests for the exp and fft algorithms
+        for algo in [self.algoExp, self.algoFft]:
+            with self.subTest(algo=algo):
```
teweitsai: I think you could consider to refactor this function to reduce the layers of indentation. There are 5 layers now. If it is not too hard, I would recommend the 3 indentation layers as the maximum. If you have too many layers, it would be hard to maintain.

teweitsai (approved):
This PR looks good to me. I think you could wait for Bryce's and Chris's comments as well. There are some coding style problem I think you should have the consensus with them.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Fixed

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: I think it makes sense to use a `getHistory` method here since the rest of this class uses `get` functions. Then the entire class can be converted to the pythonic style at the same time, rather than having two different conventions in the same class.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: I added a note to the `_recordItem()` docstring that if you record any new items in the algorithm history, you should add that item to the `getHistory()` docstring. Do you think that is sufficient?

I added units for the zernikes. I don't think the wavefronts have units? If they do, I haven't been able to find the.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: In python, a function should either mutate something, or return something, but _not both_. Having a function return objects that it also mutates is misleading and leads to hard-to-debug errors. 

We can make this problem more obvious by writing

```python
I1_masked, I2_masked = self._applyI1I2mask_pupil(I1, I2)
```

This is misleading for two reasons:
1. It suggests that the original objects, `I1` and `I2`, are unchanged. However, this is not the case. If you inspect `I1` and `I2`, you will find that their image arrays have been masked.
2. It suggests that `I1_masked` and `I2_masked` are new objects, but again, this is not the case. `I1_masked` is just a new pointer to _the exact same object_ that `I1` points to. Changing `I1_masked` also changes `I1`, and vice versa. The same is true for `I2_masked` and `I2`.

In summary, since `self._applyI1I2mask_pupil()` mutates `I1` and `I2`, it should not return them.

python/lsst/ts/wep/cwfs/Algorithm.py
teweitsai: The adding of note works to me.

teweitsai: I think the unit of wavefront is `nm` or `um`, which depends on the unit you want to hold.

python/lsst/ts/wep/cwfs/Algorithm.py
teweitsai: As long as you have the consensus with Bryce, any way works to me.

python/lsst/ts/wep/cwfs/Algorithm.py
teweitsai: In this case, you can add the doc string in `self._applyI1I2mask_pupil()` to say the inputs will be mutated.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: I think you mean using `dI` and `I0` for `num` and `den`, instead of directly calculating from `I1image` and `I2image`?

I made this change to reduce code duplication: the "exp" algorithm also calculates `dI` and `I0`. Both algorithms also have to check which image is intra/extrafocal to make sure the sign of the expression is correct (this change was necessary to restore the symmetry of I1 and I2). Having both algorithms use the same function for this is easier to maintain and read, rather than having two slightly different copies in two different locations, which each have to be maintained separately.

I think it is still straightforward for the math, since it keeps everything in terms of `dI` and `I0` which are the terms used in the original math expression `# Wavefront signal S=-(1/I0)*(dI/dz)`

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done!

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Resolved in conversation above

tests/cwfs/test_algorithm.py
jfcrenshaw: Yeah I was worried about this... there are two problems:

1. `unittest` requires adding two layers of indentation to run the same tests on two different methods:

```python
# we will run all history tests for the exp and fft algorithms
for algo in [self.algoExp, self.algoFft]:
    with self.subTest(algo=algo):
```

I could remove this and duplicate the test code for each algorithm. That might be more readable, but would mean we have multiple copies of the same tests that have to be maintained in parallel, which is a headache.

Alternatively, I could remove this and avoid most of the code duplication by not running all of the tests for each version of the algorithm. This is probably fine, since most of the logging happens in code that is common to both algorithm versions.

2. the tests require nested for-loops because they have to traverse and check the contents of nested objects. I think these loops are harder to get around.

So I could get down to three layers of indentation by eliminating problem 1. I think not running all of the tests for each algorithm version is better than just duplicating all of the test code for each. What do you think?

tests/cwfs/test_algorithm.py
jfcrenshaw: Ah yes, thank you! I should have forced the pre-implementation test to fail instead of just passing! Working on this now.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done

tests/cwfs/test_algorithm.py
teweitsai: Under this condition, I would prefer to stay at the original layers of indentation to make sure we could test all the conditions. If we could come out a better realization to test all the conditions with the lower indented layers in a latter time, we could do a PR at that time. Thanks!

python/lsst/ts/wep/cwfs/Algorithm.py
teweitsai: If this is the case, it is fine to me.

tests/cwfs/test_algorithm.py
jmeyers314: Can you dedent using a helper method?

```python
for algo in [self.algoExp, self.algoFft]: 
    with self.subTest(algo=algo):
        self.checkAlgHistory(algo)
```

jfcrenshaw: Ah yes that is a smart and in hindsight incredibly obvious solution 🤦‍♂️

tests/cwfs/test_algorithm.py
jfcrenshaw: Done.

tests/cwfs/test_algorithm.py
jfcrenshaw: Done. I used the same structure for testing the I1 <-> I2 symmetry for both algorithms

python/lsst/ts/wep/cwfs/Algorithm.py
suberlak: Did @jbkalmbach  weigh in ?  Does it affect adding `get` to any other elements of `Algorithm` ?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -493,6 +502,100 @@ def _getWavefrontMapWithMaskApplied(self, wfMap):
...
+            - initS - the initial "signal"
+            - FFT - the Fast Fourier transform of initS
+            - estWf - the estimated wavefront
+            - estS - the estimated signal
+
+        So for example, if you want to get the estimated signal from the 3rd
+        step of the innerloop during the 2nd step of the outer loop, you
+        would do the following:
+        history = algorithm.getHistory()
+        signal = history[2]["innerLoop"][3]["estS"]
```
suberlak: If the `history` dic starts with 0, isn't ` history[2]` for 3rd step of the outer loop?

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -548,7 +651,17 @@ def runIt(self, I1, I2, model, tol=1e-3):
         tol : float, optional
             Tolerance of difference of coefficients of Zk polynomials compared
             with the previous iteration. (the default is 1e-3.)
```
suberlak: I'd add `iteration of the outer loop` to be more clear

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -739,24 +857,43 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
...
+            # Record the pupil mask that was used in the previous function,
+            # plus the newly masked images
+            self._recordItem(self.mask_pupil.copy(), "pupilMask", jj, debugLevel=1)
+            self._recordItem(I1.getImg().copy(), "maskedI1", jj, debugLevel=1)
+            self._recordItem(I2.getImg().copy(), "maskedI2", jj, debugLevel=1)
 
             # Solve the Poisson's equation
             self.zc, self.West = self._solvePoissonEq(I1, I2, jj)
 
+            # Record the residual zernikes (in nm) and wavefront
```
suberlak: I think `Zernike coefficients`  is used elsewhere throughout `ts_wep`, just like "Legendre polynomials" or "Bessels functions"   - this is because Fritz Zernike is a name https://telescope-optics.net/zernike_aberrations.htm

python/lsst/ts/wep/cwfs/Algorithm.py
```diff
@@ -769,6 +906,12 @@ def _singleItr(self, I1, I2, model, tol=1e-3):
                     self.getObsOfZernikes(),
                 )
 
+                # Record the total zernikes (in nm) and wavefront
```
suberlak: Same as above - `Zernike coefficients` or `combined coefficients of Zernike polynomials`...

python/lsst/ts/wep/cwfs/CompensableImage.py
suberlak: thanks for the clarification! Would be useful to add that info somewhere in the docs (README ?) so it doesn't get lost in the graveyard of closed PRs and deleted branches

suberlak (changes requested):

python/lsst/ts/wep/cwfs/Algorithm.py
jbkalmbach: Sure, it makes sense to just convert everything over at the same time. But I would agree with Te-Wei that if you are going to use a `get` function then rename `self.history` to `self._history` to imply that it should only be for internal use.

python/lsst/ts/wep/cwfs/CompensableImage.py
jfcrenshaw: @suberlak I added a section about this to the bottom of the README. Would you mind reading it and letting me know if you think it needs any changes?

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Nice catch! Fixed.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

python/lsst/ts/wep/cwfs/Algorithm.py
jfcrenshaw: Done.

README.md
```diff
@@ -161,6 +161,25 @@ file [`data/lsst/focalPlaneLayout.txt`](https://github.com/lsst-ts/phosim_syseng
 when phosim creates images and the 180 degree difference offsets that and creates the correct orientation with the rest of the focal plane.
 This is shown in the notebook [here](https://github.com/suberlak/AOS/blob/main/AOS_DM-30367_summary.ipynb).
 
+## Note on the orientation of images and masks
+
+***tl;dr***
```
suberlak: small thing, but I think that's too informal for `ts` package if you ask me

README.md
```diff
@@ -161,6 +161,25 @@ file [`data/lsst/focalPlaneLayout.txt`](https://github.com/lsst-ts/phosim_syseng
...
+***Image compensation puts both images into the same orientation.***
+***Pupil masks are generated in the same orientation.***
+***Therefore, we do not need to explicitly rotate images or masks during the algorithm.***
```
suberlak: Perhaps I'm not the best person to judge, but run "rotate ... during the algorithm" by someone else - does it seem odd just to me? Seems like a verb is missing or rephrase needed eg. "rotate ... during the execution of the algorithm" or "rotate... inside the algorithm..."

README.md
```diff
@@ -161,6 +161,25 @@ file [`data/lsst/focalPlaneLayout.txt`](https://github.com/lsst-ts/phosim_syseng
...
+
+The Zernike estimation algorithm uses donut images and masks to solve the transport of intensity equation (TIE; see [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X)).
+A pair of intra- and extra-focal donuts is used to approximate the gradient of the donut intensity with respect to the focal direction (i.e. the left side of [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equation 1), while the average of the pair is used to approximate the in-focus image (i.e. $I$ on the right side of [Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equation 1).
```
suberlak: MNRAS always preferred "right hand side", "left hand side" to just "left side"... Is it just BE vs AE or an actual style improvement?

README.md
```diff
@@ -161,6 +161,25 @@ file [`data/lsst/focalPlaneLayout.txt`](https://github.com/lsst-ts/phosim_syseng
...
+
+In order to do this, we must ensure that the donuts are in the same orientation, and apply a mask that corrects for potential vignetting of the intrafocal donut.
+Since the two donuts in the pair are on opposite sides of focus, their orientations are initially rotated by 180 degrees with respect to each other.
+However, before calculating the mean and difference intensities, the donuts are mapped to the pupil plane and "compensated" for the current best-guess of the optical aberations ([Xin 2015](https://ui.adsabs.harvard.edu/abs/2015ApOpt..54.9045X), Equations 9-11).
```
suberlak: "aberrations" (typo)

suberlak (changes requested):
Thanks for updating the README!

README.md
jfcrenshaw: Okay, I changed this to "Summary"!

README.md
jfcrenshaw: Changed this to "inside the algorithm".

README.md
jfcrenshaw: I googled this, and they're usually interchangeable, but apparently it's convention in mathematics in use "right-hand side" and "left-hand side" because this matches the commonly-used abbreviations "RHS" and "LHS".

So I changed this to "right-hand side", etc.

README.md
jfcrenshaw: 🙏

suberlak (approved):
Already reviewed the README changes, thanks for improvement!

PR #153, opened 2022-10-25
Update test_generateDonutCatalogWcsTask to work with more recent vers…

tribeiro:
…ions of the DM stack.


teweitsai (approved):
Looks good!

PR #154, opened 2022-11-01
Tickets/DM-36304: remove imageCoCenter

suberlak:
Given the results of https://jira.lsstcorp.org/browse/DM-36714  and https://jira.lsstcorp.org/browse/DM-36304 , I open this PR to remove the `imageCoCenter` step, given that it makes no difference on the result of the calculation, and therefore saves extra computational time. To summarize these tickets, it has been found that the amount of shift imparted by `imageCoCenter` , being by design `<6` pixels (assuming defocal offset of `1.5` mm and pixels size `10` microns), makes no difference on the recovered Zernikes. This is because this minute shift is counteracted for in the `compensate` step where images are  transformed to the pupil plane, and `recentered`, to match the center of the projection template to the center of each donut. Simulations showed that the shift imparted  by `imageCoCenter` would have to exceed `~9` pixels to make any difference to the results of running the CWFS algorithm.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,15 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-3.1.4:
+
+-------------
+3.1.4
+-------------
+
+* Remove imageCoCenter step from Algorithm.
+* Add DeprecationWarning that imageCoCenter function in CompensableImage will be removed after January 2023.
+
 .. _lsst.ts.wep-3.1.3:
 
 -------------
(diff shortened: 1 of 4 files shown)


jbkalmbach:
Ok, sorry about the wait. I've verified that this does not affect my masked deblending results on comcam either.

jbkalmbach (dismissed):
Removing this step from `Algorithm.py` is good with me, but check with @teweitsai about the proper way to deprecate a function with a warning that it will be deleted at a later date rather than deleting it all at once from `CompensableImage.py`.

jbkalmbach:
See here for a deprecationWarning example: https://github.com/lsst-ts/ts_wep/blob/d41a81fe380c09432950d0db933e7b664b8782ea/python/lsst/ts/wep/cwfs/Tool.py#L64-L68

teweitsai (approved):
Thanks for this improvement and deprecation warning to let people have time to update the code!

PR #155, opened 2022-11-22
Add default value to DonutStamp for DFC_DIST.

jbkalmbach:
Added default value to DonutStamp for DFC_DIST to allow the butler to read DonutStamp from repositories created with older versions of ts_wep.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-3.1.3:
+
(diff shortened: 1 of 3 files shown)


suberlak (approved):
I've tested with two older repos, including `/sdf/group/rubin/ncsa-project/project/aos/vignetting_tests/vignette_test_no_rot/phosimData/`. Now they can be read with butler, including eg. `donutStampsExtra`.    Thanks for the update! Probably a rebase will require an update to version history.

PR #157, opened 2022-11-30
Tickets/DM-37160: update corner sensor test files 

suberlak:


teweitsai (approved):
Thanks to the quick fix of test files!

PR #159, opened 2022-12-09
tickets/DM-37289: Port Latiss functionality into new pipeline and deprecate EstimateZernikesTasks

jbkalmbach:


jbkalmbach:
Adding @suberlak as review to make sure Latiss functionality all moved over.

teweitsai:
I felt the `CutOutDonutsLatissTask.py` is a duplicated code. Is it possible to update the `CutOutDonutsTask` to use the configuration to identify the different instrument and provide the related business logic? Or this will increase the code complexity?

jbkalmbach:
> I felt the `CutOutDonutsLatissTask.py` is a duplicated code. Is it possible to update the `CutOutDonutsTask` to use the configuration to identify the different instrument and provide the related business logic? Or this will increase the code complexity?

Hi @teweitsai. I think I addressed your concern by moving the LATISS functionality in to `CutOutDonutsScienceSensorTask` instead of making a brand new task since as you said a lot of the functionality was already there.

tests/testData/pipelineConfigs/testCutoutsLatissPipeline.yaml
```diff
@@ -0,0 +1,50 @@
...
+      doInterpolate: False
+      doBrighterFatter: False
+      doDark: False
+      doFlat: False
+      doApplyGains: True
+      doFringe: False
+      doOverscan: True
+      python: OverscanCorrectionTask.ConfigClass.fitType = 'MEDIAN'
+  generateDonutDirectDetectTask:
+    class: lsst.ts.wep.task.GenerateDonutDirectDetectTask.GenerateDonutDirectDetectTask
```
suberlak, line 33: for LATISS I've used before 'MEDIAN_PER_ROW' https://github.com/lsst-ts/ts_wep/blob/96ffe9be7795e9dd6d3627880e17b49a9dbea414/tests/testData/pipelineConfigs/testLatissPipeline.yaml#L24   . Why is this different?  Have you tested that it works better?

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
```diff
@@ -125,23 +130,47 @@ def assignExtraIntraIdx(self, focusZVal0, focusZVal1):
...
+                intraExpIdx = 1
+            else:
+                # Need to be defocal images ('FOCUSZ != 0')
                 raise ValueError(errorStr)
-            extraExpIdx = 1
-            intraExpIdx = 0
-        elif focusZVal0 > 0:
-            # Check that other image does not have same defocal direction
-            if focusZVal1 >= 0:
+        elif cameraName == "LATISS":
```
suberlak, line 154: It appears to be a port of https://github.com/lsst-ts/ts_wep/blob/96ffe9be7795e9dd6d3627880e17b49a9dbea414/python/lsst/ts/wep/task/EstimateZernikesLatissTask.py#L112  , yes?

policy/cwfs/instData/auxTel/instParamCalcZernikesConfig.yaml
```diff
@@ -0,0 +1,34 @@
+description: LATISS Telescope Parameters
```
suberlak: It's such a long code reorganization that it is a but difficult to reverse-engineer to figure out what was the intention (the ticket https://jira.lsstcorp.org/browse/DM-37289 doesn't provide any details). By the look of files I'm guessing that some functionality of `EstimateZernikes` is split into two tasks, `cutOutDonuts` and `calcZernikes`  .  By looking at files it seems that some configs are a copy-paste from another config. I would find it very helpful to see the executive summary of intended changes + a trace of which configs went from where to where.  Since majority of the code seems to be moved rather than written from the scratch, it would help also find out what's new.

policy/cwfs/instData/auxTel/instParamCalcZernikesConfig.yaml
```diff
@@ -0,0 +1,34 @@
...
+instrument: lsst.obs.lsst.Latiss
+
+tasks:
+  CutOutDonutsScienceSensorTask:
+    class: lsst.ts.wep.task.CutOutDonutsScienceSensorTask.CutOutDonutsScienceSensorTask
+    config:
+      # Obscuration (inner_radius / outer_radius of primary mirror)
+      instObscuration: 0.3525
```
suberlak: For instance this data seems to be related to https://github.com/lsst-ts/ts_wep/blob/develop/policy/cwfs/instData/auxTel/instParamPipeConfig.yaml  . Is the `instParamPipeConfig` being removed? Are we duplicating any configs anywhere (because if yes, then any change to any parameter would have to be propagated everywhere, would be good to have one place for a parameter that is intended to be describing the instrument).

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
```diff
@@ -91,17 +91,19 @@ def runQuantum(
...
         donutCats = butlerQC.get(inputRefs.donutCatalog)
-        camera = butlerQC.get(inputRefs.camera)
 
         outputs = self.run(exposures, donutCats, camera)
 
         butlerQC.put(outputs.donutStampsExtra, outputRefs.donutStampsExtra[extraIdx])
         butlerQC.put(outputs.donutStampsIntra, outputRefs.donutStampsIntra[extraIdx])
 
-    def assignExtraIntraIdx(self, focusZVal0, focusZVal1):
+    def assignExtraIntraIdx(self, focusZVal0, focusZVal1, cameraName):
```
suberlak, line 106: Is this a combination of what was hitherto in https://github.com/lsst-ts/ts_wep/blob/96ffe9be7795e9dd6d3627880e17b49a9dbea414/python/lsst/ts/wep/task/EstimateZernikesLatissTask.py#L112  and https://github.com/lsst-ts/ts_wep/blob/96ffe9be7795e9dd6d3627880e17b49a9dbea414/python/lsst/ts/wep/task/EstimateZernikesScienceSensorTask.py#L93 ?

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
```diff
@@ -125,23 +130,47 @@ def assignExtraIntraIdx(self, focusZVal0, focusZVal1):
...
+                extraExpIdx = expDicIdx[extraExp]
+            else:
+                # Need two different focusz values
                 raise ValueError(errorStr)
-            extraExpIdx = 0
-            intraExpIdx = 1
         else:
-            # Need to be defocal images ('FOCUSZ != 0')
+            errorStr = str(
+                "Invalid cameraName parameter: {cameraName}. Camera must  "
```
suberlak: Did you check that the error string provides correct error ? The f-string appears to be missing the f....   ( `f"Invalid...{cameraName}")`

tests/task/test_cutOutDonutsScienceSensorTask.py
```diff
@@ -107,39 +107,72 @@ def testValidateConfigs(self):
...
             str(context.exception),
         )
 
+    def testAssignExtraIntraIdxInvalidCamera(self):
+
+        cameraName = "WrongCam"
+        with self.assertRaises(ValueError) as context:
+            self.task.assignExtraIntraIdx(-1, 1, cameraName)
+        errorStr = str(
+            "Invalid cameraName parameter: {cameraName}. Camera must  "
```
suberlak: Does this f-string miss the `f` as above?  Does it pass the appropriate (erroneous) `{cameraName}` ?

tests/task/test_cutOutDonutsLatissTask.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        testIntraStamps = self.task.cutOutStamps(
+            exposureIntra, donutCatalogIntra, DefocalType.Intra, camera.getName()
+        )
+
+        for donutStamp, cutOutStamp in zip(taskOut.donutStampsExtra, testExtraStamps):
+            self.assertMaskedImagesAlmostEqual(
+                donutStamp.stamp_im, cutOutStamp.stamp_im
+            )
+        for donutStamp, cutOutStamp in zip(taskOut.donutStampsIntra, testIntraStamps):
```
suberlak, line 207: If I understand correctly this checks if donuts are the same when the task is run as a pipetask with `run` or whether the `cutOutStamps` is called from the `task` itself. But it doesn't check if any donuts are actually there...

tests/task/test_cutOutDonutsLatissTask.py
```diff
@@ -0,0 +1,200 @@
...
+        self.assertEqual(intraIdx, 0)
+
+        with self.assertRaises(ValueError):
+            self.task.assignExtraIntraIdx(focusZextra, focusZextra, "LATISS")
+        with self.assertRaises(ValueError) as context:
+            self.task.assignExtraIntraIdx(focusZintra, focusZintra, "LATISS")
+        self.assertEqual(
+            "Must have two images with different FOCUSZ parameter.",
+            str(context.exception),
+        )
```
suberlak, line 145: Perhaps here you want to add testing whether `task.assignExtraIntraIdx` throws a correct error when eg. `Hubble`  is passed as `instrument` ?

suberlak: Just like below in `testAssignExtraIntraIdxInvalidCamera` in `test_cutOutDonutsScienceSensorTask` ?

tests/task/test_generateDonutDirectDetectTask.py
```diff
@@ -154,13 +154,13 @@ def testPipeline(self):
...
 
-        result_y = np.sort(outputDf['centroid_y'].values)
+        result_y = np.sort(outputDf["centroid_y"].values)
         truth_y = np.sort(np.array([3196, 2198, 2196, 3197]))
-        diff_y = np.sum(result_y-truth_y)
+        diff_y = np.sum(result_y - truth_y)
 
-        result_x = np.sort(outputDf['centroid_x'].values)
+        result_x = np.sort(outputDf["centroid_x"].values)
         truth_x = np.sort(np.array([3815, 2814, 2811, 3812]))
```
suberlak: Is this the result of running yet another version of black?  I wonder if it could be a separate PR (unless it fails CI)

tests/testData/pipelineConfigs/testCalcZernikesLatissPipeline.yaml
```diff
@@ -0,0 +1,53 @@
...
+      doDefect: False
+      doNanMasking: False
+      doInterpolate: False
+      doBrighterFatter: False
+      doDark: False
+      doFlat: False
+      doApplyGains: True
+      doFringe: False
+      doOverscan: True
+      python: OverscanCorrectionTask.ConfigClass.fitType = 'MEDIAN'
```
suberlak: Is this using parts of https://github.com/lsst-ts/ts_wep/blob/develop/tests/testData/pipelineConfigs/testLatissPipeline.yaml ? If so, the `fitType` that worked before was `MEDIAN_PER_ROW`. Did you see better postISR image quality with `MEDIAN` ?

suberlak (changes requested):
Thanks for the improvement!

tests/testData/pipelineConfigs/testCutoutsLatissPipeline.yaml
jbkalmbach: No, I think I just copied over the FAM values and didn't notice the difference in the LATISS task. I'll update that.

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
jbkalmbach: Yes

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
jbkalmbach: Yes, as a way to reduce duplication as mentioned by Te-Wei above.

tests/testData/pipelineConfigs/testCalcZernikesLatissPipeline.yaml
jbkalmbach: No, just the same thing as the cutouts config. Updated.

tests/task/test_generateDonutDirectDetectTask.py
jbkalmbach: Sure, this can wait for another PR

python/lsst/ts/wep/task/CutOutDonutsScienceSensorTask.py
jbkalmbach: Looks like the test and this are both missing the f designation so they were passing but with the wrong message. Fixed now.

tests/task/test_cutOutDonutsLatissTask.py
jbkalmbach: I don't think that's necessary because it uses the same function so I would just be testing the same behavior on the same piece of code.

tests/task/test_cutOutDonutsLatissTask.py
jbkalmbach: Adding in some tests on the donut catalog content.

policy/cwfs/instData/auxTel/instParamCalcZernikesConfig.yaml
jbkalmbach: There actually wasn't that much capability that needed to be ported into the new pipelines. The only thing was adding the additional `assignExtraIntraIdx` functionality into `CutOutDonutsScienceSensor`. 

Almost all of the new code is adding tests specifically for the LATISS based upon the `test_estimateZernikesLatissTask` tests you originally wrote.

We already have documentation on the `CutOutDonuts` and `CalcZernikes` tasks in notebooks here: https://github.com/lsst-ts/ts_analysis_notebooks/tree/develop/aos/closed_loop that show how to use them.

I'll answer questions about configs in response to your comment below.

policy/cwfs/instData/auxTel/instParamCalcZernikesConfig.yaml
jbkalmbach: `instParamPipeConfig` only works with the `EstimateZernikesTask` family of tasks (you can see the tasks prescribed by that yaml). This file is the equivalent but using the new pipeline tasks. When we remove the `EstimateZernikes` tasks in the future we can change the `instParamPipeConfig` files to use the `CutOutDonuts` and `CalcZernikes` tasks as this new file here does. Or if you think it would be better just to switch everything over now I could do that?

teweitsai: I think we do begin to have the duplication of configuration files.
In the TS control systems, we have the default configuration file and the overwrite file.
For example: https://github.com/lsst-ts/ts_config_mttcs/tree/develop/MTM2/v2

The default is `_init.yaml`.
The content in `_summit.yaml` will overwrite the existed setting in `_init.yaml`.
I think we could have a similar thing for the `ts_wep` to minimize the duplication of configuration file.
It should be easier in future if we need to remove a task and add others new.

BTW, do we still need this:
https://github.com/lsst-ts/ts_wep/blob/tickets/DM-37289/policy/cwfs/donutTemplateData/phosimTemplates/.gitkeep

jbkalmbach: Ok, since I'm deprecating the `EstimateZernikes` I will just replace the default configuration files with the new pipeline and we will only have one file for each instrument.

And yes, I think we should keep that for now since we still have the functionality to add phosimTemplates but we just haven't been using them. Maybe in the future we can test batoid templates and then decide what to do with that.

suberlak (approved):

PR #160, opened 2022-12-16
Tickets/DM-35648: opticalModel for auxTel can't be offAxis 

suberlak:
Add a `ValueError` to `WfEstimator` config if `auxTel`  is trying to use `offAxis` optical model.

### DIFF
--- a/python/lsst/ts/wep/WfEstimator.py
+++ b/python/lsst/ts/wep/WfEstimator.py
@@ -171,12 +171,16 @@ class WfEstimator(object):
         """
 
         if solver not in ("exp", "fft"):
-            raise ValueError("Poisson solver can not be '%s'." % solver)
+            raise ValueError("Poisson solver cannot be '%s'." % solver)
 
         if opticalModel not in ("paraxial", "onAxis", "offAxis"):
-            raise ValueError("Optical model can not be '%s'." % opticalModel)
+            raise ValueError("Optical model cannot be '%s'." % opticalModel)
         else:
             self.opticalModel = opticalModel
+
+        if camType == CamType.AuxTel:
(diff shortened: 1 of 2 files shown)


python/lsst/ts/wep/WfEstimator.py
```diff
@@ -171,13 +171,17 @@ def config(
...
 
         if opticalModel not in ("paraxial", "onAxis", "offAxis"):
-            raise ValueError("Optical model can not be '%s'." % opticalModel)
+            raise ValueError("Optical model cannot be '%s'." % opticalModel)
         else:
             self.opticalModel = opticalModel
 
+        if camType == CamType.AuxTel:
+            if opticalModel not in ("paraxial", "onAxis"):
+                raise ValueError("Optical model cannot be '%s' for AuxTel." % opticalModel)
```
teweitsai: Please help to add the description of  this new `ValueError` in function doc string. Thanks!

teweitsai (approved):
Looks good. Thanks!

python/lsst/ts/wep/WfEstimator.py
jfcrenshaw: You could decrease the number of indentations by combining both if-statements into a single if-statement. Also the official python docs prefer f-strings over modulo strings, so I'd recommend changing these lines to:
```
if camType == CamType.AuxTel and opticalModel not in ("paraxial", "onAxis"):
    raise ValueError(f"Optical model cannot be {opticalModel} for AuxTel.")
```
(but these are just style changes, not substantive changes, so feel free to ignore me!)

suberlak: Thank you both, I've implemented these comments to improve this PR

PR #161, opened 2023-01-12
Tickets/dm 37355

jbkalmbach:


policy/task/magLimitStar.yaml
```diff
@@ -9,26 +9,26 @@
 # The values of dark sky in the following link is used (commit: 91ed8f6):
 # https://github.com/bxin/scratchBook/blob/master/saturation_limit/saturation_limit_donut.ipynb
 
-filterU:
```
suberlak: I think this change of name will break `getMagLimits` in `ts_phosim`, which reads filter magnitude limits directly from this file: https://github.com/lsst-ts/ts_phosim/blob/d41cd1429d6a3efc138222ee3971bc33e4ca55aa/python/lsst/ts/phosim/CloseLoopTask.py#L489 is it necessary?

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -204,6 +205,10 @@ def donutCatalogToDataFrame(
             ReferenceObjectLoader search over the detector footprint.
             If None then it will return an empty dataframe.
             (the default is None.)
+        filterName
+        blendCentersX
+        blendCentersY
```
teweitsai: doc string?

policy/task/magLimitStar.yaml
jbkalmbach: No, I can change it back. This is just my hacking to get things working on a first draft. Thanks for letting me know about that.

python/lsst/ts/wep/cwfs/Instrument.py
```diff
@@ -170,15 +170,19 @@ def configFromFile(
...
+            + "See examples in policy/cwfs/instData."
+        )
         for taskDict in instParamReader.getContent()["tasks"].values():
-            if "instObscuration" in list(taskDict["config"].keys()):
-                instConfigDict = taskDict["config"]
+            try:
+                if "instObscuration" in list(taskDict["config"].keys()):
+                    instConfigDict = taskDict["config"]
+            except KeyError:
+                continue
```
teweitsai: Is it possible to do:

```
keys = list(taskDict["config"].keys())
if (len(keys) != 0) and ("instObscuration" in keys):
    instConfigDict = taskDict["config"]
```

By doing this, we could reduce the indentation layer and avoid the `try`.
If `(len(keys) = 0)`, the program will not do the following  `("instObscuration" in keys)`.

python/lsst/ts/wep/cwfs/Instrument.py
```diff
@@ -170,15 +170,19 @@ def configFromFile(
...
 
         if instConfigDict is None:
-            raise ValueError(
-                "Instrument configuration file does not have expected format. "
-                + "See examples in policy/cwfs/instData."
-            )
+            raise ValueError(invalidFormatMsg)
```
teweitsai, line 184: I think you could put the line 173 (`invalidFormatMsg`) just be before line 185 to make the code coherent if you do not need to reuse the  `invalidFormatMsg`.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -119,6 +126,8 @@ def run(self, sourceCat, detector):
             Catalog of sources to select from.
         detector : `lsst.afw.cameraGeom.Detector`
             Detector object from the camera.
+        filterName : str
```
teweitsai: filterName : str --> filterName : `str`

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -166,6 +175,8 @@ def selectSources(self, sourceCat, detector):
...
```
teweitsai: filterName : str --> filterName : `str`

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
```diff
@@ -187,17 +198,26 @@ def selectSources(self, sourceCat, detector):
...
+        flux = _getFieldFromCatalog(sourceCat, fluxField)
         mag = (flux * u.nJy).to_value(u.ABmag)
-        magMin = self.config.magMin
-        magMax = self.config.magMax
         minMagDiff = self.config.isolatedMagDiff
         unblendedSeparation = self.config.unblendedSeparation
         minBlendedSeparation = self.config.minBlendedSeparation
 
+        # Use user defined inputs or ts_wep defaults depending on userMagLimit.
+        if self.config.userMagLimit:
```
teweitsai: `if self.config.userMagLimit is True` ?

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -204,6 +202,19 @@ def donutCatalogToDataFrame(
             ReferenceObjectLoader search over the detector footprint.
             If None then it will return an empty dataframe.
             (the default is None.)
+        filterName : str or `None`, optional
```
teweitsai: filterName : str or `None`, optional --> filterName : `str` or `None`, optional

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -204,6 +202,19 @@ def donutCatalogToDataFrame(
...
+            Name of camera filter. If donutCatalog is not None then
+            this cannot be None. (the default is None.)
+        blendCentersX : list or `None`, optional
```
teweitsai: blendCentersX : list or `None`, optional --> blendCentersX : `list` or `None`, optional

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -204,6 +202,19 @@ def donutCatalogToDataFrame(
...
+             X pixel position of centroids for blended objects. List
+             should be the same length as the donutCatalog. If
+             blendCentersY is not None then this cannot be None. (the default
+             is None.)
+        blendCentersY : list or `None`, optional
```
teweitsai: blendCentersY : list or `None`, optional --> blendCentersY : `list` or `None`, optional

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -220,13 +231,35 @@ def donutCatalogToDataFrame(
         blendCY = []
 
         if donutCatalog is not None:
+            filterErrMsg = "If donutCatalog is not None then filterName cannot be None."
+            if filterName is None:
+                raise ValueError(filterErrMsg)
```
teweitsai, line 246: Please add the doc string for this new `ValueError`.

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -220,13 +231,35 @@ def donutCatalogToDataFrame(
...
             centroidX = donutCatalog["centroid_x"]
             centroidY = donutCatalog["centroid_y"]
-            sourceFlux = donutCatalog[f"{self.filterName}_flux"]
-            blendCX = blendCentersX
-            blendCY = blendCentersY
+            sourceFlux = donutCatalog[f"{filterName}_flux"]
+
+            if (blendCentersX is None) and (blendCentersY is None):
+                blendCX = [[]] * len(donutCatalog)
+                blendCY = [[]] * len(donutCatalog)
```
teweitsai: I am not sure I understand the intension here. For example:

```
In [12]: a = [[]] * 3
In [13]: a[0].append(1)
In [14]: a
Out[14]: [[1], [1], [1]]
```

Is the above what you are looking for?

Or the following is the case:

```
In [20]: a = list()

In [21]: for index in range(3):
    ...:     a.append(list())
    ...: 

In [22]: a[0].append(1)

In [23]: a
Out[23]: [[1], [], []]
```

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -220,13 +231,35 @@ def donutCatalogToDataFrame(
...
+            elif (type(blendCentersX) == list) and (type(blendCentersY) is list):
```
teweitsai: `elif (type(blendCentersX) is list) and (type(blendCentersY) is list)`

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -220,13 +231,35 @@ def donutCatalogToDataFrame(
...
+                lengthErrMsg = (
+                    "blendCentersX and blendCentersY need "
+                    + "to be same length as donutCatalog."
+                )
+                if (len(blendCentersX) != len(donutCatalog)) or (
+                    len(blendCentersY) != len(donutCatalog)
+                ):
+                    raise ValueError(lengthErrMsg)
```
teweitsai, line 267: Please add the doc string for this new `ValueError`.

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
```diff
@@ -220,13 +231,35 @@ def donutCatalogToDataFrame(
...
+                blendCX = blendCentersX
+                blendCY = blendCentersY
+            else:
+                blendErrMsg = (
+                    "blendCentersX and blendCentersY must be"
+                    + " both be None or both be a list."
+                )
+                raise ValueError(blendErrMsg)
```
teweitsai, line 283: Please add the doc string for this new `ValueError`.

teweitsai (approved):

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
jmeyers314: I think `isinstance` is usually preferred.  (Slightly faster for one: https://switowski.com/blog/type-vs-isinstance/).

python/lsst/ts/wep/cwfs/Instrument.py
jbkalmbach: Okay, I did something slightly different but I was still able to avoid using the `try` statement and the extra indentation. The problem that I needed to solve was that sometimes there is no `"config"` in the taskDict.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
jbkalmbach: I think the python standard is to not include `is True`.

python/lsst/ts/wep/task/GenerateDonutCatalogWcsTask.py
jbkalmbach: It should be the second case. I've rewritten it to be like that and added a test to ensure the desired functionality.

python/lsst/ts/wep/task/DonutSourceSelectorTask.py
teweitsai: In this case, I would prefer to have some variable name like this:

```
if self.config.isUserMagLimit
```

or 

```
if self.config.useMagLimit
```

or something similar to express the idea that this variable is clearly to be `True` or `False` if possible.

jbkalmbach: That sounds good. How about `useCustomMagLimit`?

teweitsai: This works to me. Thanks!

PR #162, opened 2023-01-17
Tickets/DM-37593: update UPS table

suberlak:
Removing Gen2 `daf_persistence` requirement from the UPS table.


teweitsai (approved):
Looks good. Thanks for the quick fix!

PR #164, opened 2023-01-19
Add blend_centroid_x and blend_centroid_y to GenerateDonutDirectDetec…

jbkalmbach:
…tTask catalogs.

### DIFF
--- a/python/lsst/ts/wep/task/GenerateDonutDirectDetectTask.py
+++ b/python/lsst/ts/wep/task/GenerateDonutDirectDetectTask.py
@@ -211,6 +211,12 @@ class GenerateDonutDirectDetectTask(pipeBase.PipelineTask):
         donutCat = donutCat.rename(
             columns={"y_center": "centroid_x", "x_center": "centroid_y"}
         )
+        donutCat = donutCat.rename(
+            columns={
+                "y_blend_center": "blend_centroid_x",
+                "x_blend_center": "blend_centroid_y",
+            }
+        )
 
         # pass the boresight ra, dec
         wcs = exposure.getWcs()
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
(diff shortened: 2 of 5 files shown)


python/lsst/ts/wep/task/GenerateDonutDirectDetectTask.py
```diff
@@ -211,6 +211,12 @@ def updateDonutCatalog(self, donutCat, exposure):
         donutCat = donutCat.rename(
             columns={"y_center": "centroid_x", "x_center": "centroid_y"}
         )
+        donutCat = donutCat.rename(
```
suberlak, line 214: I think given that change it would be good to update the test for `updateDonutCatalog` in https://github.com/lsst-ts/ts_wep/blob/22a3975b1176dea512882920c278fe969e04d740/tests/task/test_generateDonutDirectDetectTask.py#L84-L86  , testing that those new columns are there.

suberlak (approved):
Just a small comment about updating the `test_generateDonutDirectDetect`, otherwise looks good! Thanks for the update!

python/lsst/ts/wep/task/GenerateDonutDirectDetectTask.py
suberlak: I also notice that in CI `test_generateDonutDirectDetectTask.py`  fails 

```
=========================== short test summary info ============================
FAILED tests/task/test_generateDonutDirectDetectTask.py::TestGenerateDonutDirectDetectTask::testPipeline - AssertionError: Element counts were not equal:
First has 1, Second has 0:  'blend_centroid_y'
First has 1, Second has 0:  'blend_centroid_x'  
```

https://tssw-ci.lsst.org/job/LSST_Telescope-and-Site/job/ts_wep/job/tickets%252FDM-37638/1/consoleFull  so that's another thing to update in that test given that now there are more elements in the donut catalog.

jbkalmbach: Added in those changes.

PR #167, opened 2023-02-02
tickets/DM-37821: Use nan string instead of None in DonutStamps.py 

jbkalmbach:
Use nans instead of None so we can convert to float and use writeFits method in DonutStamps successfully and save in butler.

Update versionHistory.


teweitsai (approved):
Looks good. Note the Jenkins test failed.

PR #168, opened 2023-02-07
Add flux sort into GenerateDonutDirectDetectTask to make it consistent with other catalog generation tasks.

jbkalmbach:


suberlak (approved):
Thanks for quick update!

PR #169, opened 2023-02-15
Tickets/DM-37903: add transposeImages config to CalcZernikesTask

suberlak:
Given the findings of https://jira.lsstcorp.org/browse/DM-37903, it is recommended not to transpose images for auxTel (and possibly other real-telescope images). This PR adds transposing images as an optional config. By default (eg. using phoSim images) the transpose is done as before, so that no tests need amendment.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-4.2.3:
+
+-------------
+4.2.3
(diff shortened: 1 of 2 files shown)


teweitsai (approved):
Looks good. Thanks for the quick fix! Please help to check the fail on Jenkins. Thanks

suberlak:
Fixed the CI issue (it was a typo in intra/extra images).  Merging.

PR #170, opened 2023-02-24
Add bandpass information into DonutStamp objects.

jbkalmbach:


python/lsst/ts/wep/task/CutOutDonutsBase.py
```diff
@@ -381,6 +381,7 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         detectorType = exposure.getDetector().getType()
         pixelScale = exposure.getWcs().getPixelScale().asArcseconds()
         camType = getCamTypeFromButlerName(cameraName, detectorType)
+        bandpass = exposure.filter.bandLabel
         # Get template
```
suberlak, line 385: Is `bandLabel` property something that has been around for some time, or was it introduced in some recent weekly? Just making sure it doesn't break when trying to run on an older dataset.

python/lsst/ts/wep/task/DonutStamp.py
```diff
@@ -61,6 +61,8 @@ class DonutStamp(AbstractStamp):
     cam_name : `str`
         Camera name for the stamp image. "LSSTCam" or "LSSTComCam"
         are available camera names currently.
+    bandpass : `str`
+        The bandpass for the stamp image.
```
suberlak, line 65: Are there any allowed/default values?

suberlak (approved):
Thanks for the update!

python/lsst/ts/wep/task/CutOutDonutsBase.py
jbkalmbach: A couple years old (looks like w_2020_50) so hopefully that shouldn't break anything you're currently using?

jbkalmbach: https://community.lsst.org/t/afw-image-filter-is-now-deprecated/4675

python/lsst/ts/wep/task/DonutStamp.py
jbkalmbach: Ah, good point. This is not optional but I will add information below in the factory method pointing out the default value used when loading stamps created with an older version.

PR #173, opened 2023-03-21
Change module names to align with DM Convention

jbkalmbach:
DM Convention is to use a lowercase first letter in module names and capital for classes inside the module. This needs to be changed in `ts_wep` in order to allow it to run in batch processing mode on the USDF.


README.md
```diff
@@ -127,6 +127,20 @@ The options used above are as follows:
...
+In `ts_wep` release 6.0.0 we updated the module names to start with a lower case letter in order to
+bring them in line with LSST Data Management convention and enable batch processing of the WEP
+Gen 3 pipeline. This will lead to errors when attempting to load `donutStamps` created with older
+versions of `ts_wep`. However adding symlinks in the following way will enable loading the older `donutStamps`
+through the LSST Science Pipelines butler once again.
+
+```bash
+cd $path_of_ts_wep/python/lsst/ts/wep/task/
+ln -s donutStamp.py DonutStamp.py
+ln -s donutStamps.py DonutStamps.py
```
teweitsai, line 139: This might make the `git` workspace dirty. If so, maybe mention this in the `README.md` as well.
BTW, the above description might not follow the MARKDOWN syntax well:

"""
In `ts_wep` release 6.0.0 we updated the module names to start with a lower case letter in order to bring them in line with LSST Data Management convention and enable batch processing of the WEP Gen 3 pipeline.
"""

For the above sentence, it looks like three lines from 132-134.
Same thing for others.

jbkalmbach: Maybe we should preemptively add these paths to the git ignore?

teweitsai: In the MARKDOWN syntax, it should be:

"""
In `ts_wep` release 6.0.0 we updated the module names to start with a lower case letter in order to bring them in line with LSST Data Management convention and enable batch processing of the WEP Gen 3 pipeline.
This will lead to errors when attempting to load `donutStamps` created with older versions of `ts_wep`.
However adding symlinks in the following way will enable loading the older `donutStamps` through the LSST Science Pipelines butler once again.
"""

The above paragraph would be three sentences (lines), which would become line 132-134 instead of line 132 only.
After you push this change to the GitHub, you would see they look like a single paragraph instead of three.
Please test and verify this.
Thanks!

README.md
```diff
@@ -127,6 +127,18 @@ The options used above are as follows:
...
+
+In `ts_wep` release 6.0.0 we updated the module names to start with a lower case letter in order to bring them in line with LSST Data Management convention and enable batch processing of the WEP Gen 3 pipeline. This will lead to errors when attempting to load `donutStamps` created with older versions of `ts_wep`. However adding symlinks in the following way will enable loading the older `donutStamps` through the LSST Science Pipelines butler once again.
+
+```bash
+cd $path_of_ts_wep/python/lsst/ts/wep/task/
+ln -s donutStamp.py DonutStamp.py
+ln -s donutStamps.py DonutStamps.py
+```
+
+Note that adding these symbolic links might make the `git` workspace dirty. Please be careful not to add these file locations when making commits.
```
teweitsai: Same comment above:

"""
Note that adding these symbolic links might make the `git` workspace dirty.
Please be careful not to add these file locations when making commits.
"""

This would be line 140 and 141.

doc/versionHistory.rst
```diff
@@ -6,6 +6,15 @@
...
 ##################
 
+.. _lsst.ts.wep-6.0.0:
+
+-------------
+6.0.0
+-------------
+
+* Rename all modules to start with lowercase in order to align with DM convention.
+* Add information into documentation about how this new change breaks repositories with data prcoessed with older versions of ts_wep and how to fix it.
```
teweitsai: prcoessed --> processed

teweitsai (approved):
Looks good to me. Thanks for this improvement!

README.md
jbkalmbach: Oh, I understand now. Sorry, I misinterpreted your earlier comment. Thanks

PR #177, opened 2023-04-11
tickets/DM-68653: Add optional ability to specify filter in GenerateDonutCatalogWcsTask.

jbkalmbach:


python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
```diff
@@ -307,7 +313,10 @@ def run(
 
         detector = exposure.getDetector()
         detectorWcs = exposure.getWcs()
-        filterName = exposure.filter.bandLabel
+        if self.config.anyFilterMapsToThis is None:
+            filterName = exposure.filter.bandLabel
+        else:
+            filterName = self.config.anyFilterMapsToThis
```
teweitsai: `filterName = exposure.filter.bandLabel if self.config.anyFilterMapsToThis is None else self.config.anyFilterMapsToThis`

teweitsai (approved):
Looks good!

PR #179, opened 2023-04-18
DM-38482: Add FitWcs Task

jbkalmbach:


python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
```diff
@@ -0,0 +1,239 @@
...
+            "visit",
+            "detector",
+            "instrument",
+        ),
+        storageClass="DataFrame",
+        name="directDetectDonutCatalog",
+    )
+    astromRefCat = connectionTypes.PrerequisiteInput(
+        doc="Reference catalog to use for astrometry",
+        name="gaia_dr2_20200414",
```
suberlak: I understand this is a required connection. If in the future some other name was to be used (rather than `20200414` version), would it necessitate change in `ts_wep`? Could it take the name of the catalog from config?

tests/task/test_generateDonutDirectDetectTask.py
```diff
@@ -1,3 +1,24 @@
+# This file is part of ts_wep.
```
suberlak: Is this preamble change meant to be part of this PR?

suberlak: Because otherwise nothing seems to have changed in `generateDonutDirectDetect`...

tests/task/test_fitWcsFromDetectedTask.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Check that run doesn't already exist due to previous improper cleanup
+        butler = dafButler.Butler(cls.repoDir)
+        registry = butler.registry
+        collectionsList = list(registry.queryCollections())
+        if cls.runName in collectionsList:
+            cleanUpCmd = writeCleanUpRepoCmd(cls.repoDir, cls.runName)
+            runProgram(cleanUpCmd)
+
+        collections = "refcats/largeCatSingleCCD,LSSTCam/calib,LSSTCam/raw/all"
```
suberlak: If I understand correctly, this test is using simulated image with an associated reference catalog that was created based on the star catalog. Do you want to include some real exposure (eg. auxTel) using the real refcat (such as the required connection `gaia_dr2_20200414` ? )

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    Specify the pipeline inputs and outputs needed
+    for FitDonutWcsTask.
+    """
+
+    inputExposure = connectionTypes.Input(
+        doc="Input exposure to make measurements on",
+        dimensions=("exposure", "detector", "instrument"),
+        storageClass="Exposure",
+        name="preFitPostISRCCD",
```
suberlak: Since this input exposure needs to have a specific name `preFitPostISRCCD` to work (or could it be set in the `config` ? ), it would be good to mention the addition of this task in `README`,  and that the intended usage is: 
*  run `ISR` creating `preFitPostISRCCD`
* run the `generateDonutDirectDetect` to get donut catalog
* run `fitWcsFromDetected`  to refit for WCS
* (optional : run `generateDonutCatalogWcsTask` to get the updated donut catalog)
* cutout donuts with `cutOutDonutsScienceSensorTask`
* use `calcZernikesTask` to fit the wavefront given the donut postage stamp images

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
```diff
@@ -0,0 +1,239 @@
...
+            " This can be set for purely-astrometric catalogs (e.g. Gaia DR2) where there is only one "
+            "reasonable choice for every camera filter->refcat mapping, but not for refcats used for "
+            "photometry, which need a filterMap and/or colorterms/transmission corrections."
+        ),
+        dtype=str,
+        default=None,
+        optional=True,
+    )
+
+    # Took these defaults from atmospec/centroiding
```
suberlak: can you expand what `atmospec/centroiding` pertains to ?

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
```diff
@@ -0,0 +1,239 @@
...
+    def setDefaults(self):
+        super().setDefaults()
+        self.astromTask.wcsFitter.retarget(FitAffineWcsTask)
+        self.astromTask.referenceSelector.doMagLimit = True
```
suberlak: So if we happen to have donuts fainter than that in some edge case, and the WCS fit fails,   will it allow eg. to skip using `generateDonutCatalogWcsTask`, but proceed to the next step (`cutOutDonuts` and `calcZernikes` ) using the donut catalog from `generateDonutDirectDetect` ? So that in a sense,  the WCS fitting could be used as an extra option - "try to refit WCS, but if it fails, still use the direct detection donut locations to do wavefront estimation".

suberlak (changes requested):
Thanks for the update! If`README` could be updated with the intended use case for this task it would be very helpful.

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
jbkalmbach: Yes, the config overrides this. As an example you can see the pipeline config file in this PR tells it to use `cal_ref_cat`.

jbkalmbach: So to answer your question, no change would be necessary here.

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
jbkalmbach: Not the way it is currently designed because of the way the connections are specified as they flow through the pipeline. If we want that in the future we could try to redesign this task and `generateDonutCatalogWcsTask` so that this task would be an optional subtask within `generateDonutCatalogWcsTask`.

tests/task/test_fitWcsFromDetectedTask.py
jbkalmbach: I have a notebook ready in `ts_analysis_notebooks` that will demonstrate this on real data. But for the tests requiring a connection to `gaia_dr2_20200414` would mean you could only run it on machines where that was available and this test wouldn't be portable so it wouldn't be consistently tested on Jenkins for example.

tests/task/test_generateDonutDirectDetectTask.py
jbkalmbach: I will make it a separate PR.

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
jbkalmbach: Adding a link to the code and describing why it's relevant.

teweitsai:
Please help to update the `doc/taskClass.uml`, `doc/developer-guide/developer-guide.rst` , and other documents to reflect the new task class in this PR.

jbkalmbach:
Ok, I actually made some big changes so that this task will generate the catalogs as well. This makes the pipeline more robust since it can handle WCS fit failures by just returning the same exposure WCS and the directDetectCatalog. I also added information on the success of the WCS fit into the task metadata to make it easily available.

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
jbkalmbach: Ok, I think this should all be in the README now. And it should be simpler with the elimination of the need to add `generateDonutCatalogWcsTask` into the pipeline.

jbkalmbach:
If the changes are enough that it would be better to open a new PR let me know.

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
```diff
@@ -190,50 +209,84 @@ def formatDonutCatalog(self, catalog):
...
+            # Except RuntimeError caused when no reference catalog
+            # available for the region covered by detector
+            except RuntimeError:
+                self.log.warning(
+                    "No catalogs cover this detector. Returning empty catalog."
+                )
+                refSelection = None
+                blendCentersX = None
+                blendCentersY = None
+
```
suberlak: So if no catalogs cover the detector (i.e. WCS fit was successful both the refCat used has no sources in the requested region), will the original donut catalog from direct detection be returned, just like for a failed WCS case?

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
```diff
@@ -37,7 +37,6 @@
 import lsst.pipe.base.connectionTypes as connectionTypes
 from lsst.utils.timer import timeMethod
 from lsst.meas.algorithms import ReferenceObjectLoader, LoadReferenceObjectsConfig
-from lsst.meas.algorithms.sourceSelector import ReferenceSourceSelectorTask
```
suberlak, line 40: I wonder what makes this not needed anymore (I'm glad it isn't, but is it removed because it's a good update anyway, or is it somehow related to the WCS fit task?)

suberlak (approved):
It's more overall comment for clarity: since the intended usage would be now to run `generateDonutDirectDetect`,  then try `fitWcsFromDetected`  (which may provide a new donut catalog),  would it be be better to  eg. call the new task `generateDonutFromRefitWcs` or something like that ? (to imply that just like for `generateDonutDirectDetect` and `generateDonutCatalogWcs` , the intended outcome is a new donut catalog)

python/lsst/ts/wep/task/fitWcsFromDetectedTask.py
jbkalmbach: Good point. Let me add some code to make that happen and a test case for it.

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
jbkalmbach: Yeah, all of the selection code occurs within the `DonutSourceSelectorTask` now so it's all in one place and this isn't necessary anymore as a result.

jbkalmbach:
> It's more overall comment for clarity: since the intended usage would be now to run `generateDonutDirectDetect`, then try `fitWcsFromDetected` (which may provide a new donut catalog), would it be be better to eg. call the new task `generateDonutFromRefitWcs` or something like that ? (to imply that just like for `generateDonutDirectDetect` and `generateDonutCatalogWcs` , the intended outcome is a new donut catalog)

Yeah, that makes sense. I'll change the name to `generateDonutFromRefitWcs`.

PR #180, opened 2023-04-21
DM-38804: Import MaskedImage directly from afw.image.

jbkalmbach:


suberlak (approved):
Looks good - I searched `ts_wep` repo to double check that this PR includes all places where `afwImage.maskedImage` was mentioned. Thanks for the quick fix!

PR #182, opened 2023-04-27
DM-38924: Add license information to test_generateDonutDirectDetectTask.

jbkalmbach:


suberlak (approved):
Thanks for the update! Looks good to me

PR #183, opened 2023-05-01
DM-38963: Unified dI/dz calculation and added dI/dz and I0 to algorithm history.

jfcrenshaw:
https://jira.lsstcorp.org/browse/DM-38963

### DIFF
--- a/python/lsst/ts/wep/cwfs/algorithm.py
+++ b/python/lsst/ts/wep/cwfs/algorithm.py
@@ -574,6 +574,8 @@ class Algorithm(object):
             - pupilMask - pupil mask that is applied before zernike calculation
             - maskedI1 - the masked version of the compensated I1
             - maskedI2 - the masked version of the compensated I2
+            - dI/dz - z derivative of the image: +/- (I1 - I2) / (2 * deltaZ)
+            - I0 - the mean central image: (I1 + I2) / 2
             - residZk - the estimated residual zernikes (units: nm)
             - residWf - the estimated residual wavefront (units: nm)
             - totZk - the current best estimate of the zernikes (units: nm)
@@ -967,7 +969,8 @@ class Algorithm(object):
             Intra- or extra-focal image.
         iOutItr : int, optional
             ith number of outer loop iteration which is important in "fft"
-            algorithm. (the default is 0.)
+            algorithm, and in saving dI/dz and I0 to the algorithm history.
+            (the default is 0.)
 
         Returns
         -------
@@ -985,6 +988,13 @@ class Algorithm(object):
 
         # Calculate the differential Omega
         dOmega = aperturePixelSize**2
+
+        # Calculate partialI=dI/dz and I0
+        partialI, I0 = self._getPartialIandI0(I1, I2)
+
+        # Record the image gradient and mean central image
+        self._recordItem(partialI.copy(), "dI/dz", iOutItr, debugLevel=1)
+        self._recordItem(I0.copy(), "I0", iOutItr, debugLevel=1)
 
         # Solve the Poisson's equation based on the type of algorithm
         numTerms = self.getNumOfZernikes()
@@ -1027,7 +1037,7 @@ class Algorithm(object):
             u2v2[ctrIdx, ctrIdx] = np.inf
 
             # Calculate the wavefront signal
-            Sini = self._createSignal(I1, I2, cliplevel)
+            Sini = self._createSignal(partialI, I0, cliplevel)
 
             # Find the just-outside and just-inside indices of a ring in pixels
             # This is for the use in setting dWdn = 0
@@ -1124,15 +1134,12 @@ class Algorithm(object):
             # Use the integration method by serial expansion to solve the
             # Poisson's equation
 
-            # Calculate dI and I0
-            dI, I0 = self._getdIandI0(I1, I2)
-
             # Get the x, y coordinate in mask. The element outside mask is 0.
             xSensor, ySensor = self._inst.getSensorCoor()
             xSensor = xSensor * self.mask_comp
             ySensor = ySensor * self.mask_comp
 
-            # Create the F matrix and Zernike-related matrixes
+            # Create the vector F and matrix M
 
             # Get Zernike and gradient bases from cache.  These are each
             # (nzk, npix, npix) arrays, with the first dimension indicating
@@ -1140,8 +1147,8 @@ class Algorithm(object):
             zk, dzkdx, dzkdy = self._zernikeBasisCache()
 
             # Eqn. (19) from Xin et al., Appl. Opt. 54, 9045-9054 (2015).
-            # F_j = \int (d_z I) Z_j d_Omega
-            F = np.tensordot(dI, zk, axes=((0, 1), (1, 2))) * dOmega
+            # F_j = \int (dI/dz) Z_j d_Omega
+            F = np.tensordot(partialI, zk, axes=((0, 1), (1, 2))) * dOmega
             # Eqn. (20) from Xin et al., Appl. Opt. 54, 9045-9054 (2015).
             # M_ij = \int I (grad Z_j) . (grad Z_i) d_Omega
             #      =   \int I (dZ_i/dx) (dZ_j/dx) d_Omega
@@ -1150,14 +1157,6 @@ class Algorithm(object):
             Mij += np.einsum("ab,iab,jab->ij", I0, dzkdy, dzkdy)
             Mij *= dOmega / (self._inst.apertureDiameter / 2.0) ** 2
 
-            # Calculate dz
-            dz = (
-                2
-                * self._inst.focalLength
-                * (self._inst.focalLength - self._inst.defocalDisOffsetInM)
-                / self._inst.defocalDisOffsetInM
-            )
-
             # Define zc
             zc = np.zeros(numTerms)
 
@@ -1165,7 +1164,7 @@ class Algorithm(object):
             idx = self.getZernikeTerms()
 
             # Solve the equation: M*W = F => W = M^(-1)*F
-            zc_tmp = np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0] / dz
+            zc_tmp = np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0]
             zc[idx] = zc_tmp
 
             # Estimate the wavefront surface based on z4 - z22
@@ -1176,7 +1175,7 @@ class Algorithm(object):
 
         return zc, West
 
-    def _createSignal(self, I1, I2, cliplevel):
+    def _createSignal(self, partialI, I0, cliplevel):
         """Calculate the wavefront singal for "fft" to use in solving the
         Poisson's equation.
 
@@ -1186,10 +1185,10 @@ class Algorithm(object):
 
         Parameters
         ----------
-        I1 : CompensableImage
-            Intra- or extra-focal image.
-        I2 : CompensableImage
-            Intra- or extra-focal image.
+        partialI : numpy.ndarray
+            The partial derivative of the image with respect to z, dI/dz.
+        I0 : numpy.ndarray
+            The mean central image.
         cliplevel : float
             Parameter to determine the threshold of calculating I0.
 
@@ -1199,14 +1198,9 @@ class Algorithm(object):
             Approximated wavefront signal.
         """
 
-        # Calculate I0 and dI
-        dI, I0 = self._getdIandI0(I1, I2)
-
-        # Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be
-        # -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)
-        # we will ignore the (1/delta z) until later in this method
-        num = -dI
-        den = 2 * I0
+        # Wavefront signal S=-(1/I0)*(dI/dz)=-(dI/dz)/I0
+        num = -partialI
+        den = I0
 
         # Define the effective minimum central signal element by the threshold
         # ( I0=(I1+I2)/2 )
@@ -1222,20 +1216,12 @@ class Algorithm(object):
         # Define the effective minimum central signal element
         den[den < medianThreshold * cliplevel] = 1.5 * medianThreshold
 
-        # Calculate delta z = f(f-l)/l, f: focal length, l: defocus distance of
-        # the image planes
-        deltaZ = (
-            self._inst.focalLength
-            * (self._inst.focalLength - self._inst.defocalDisOffsetInM)
-            / self._inst.defocalDisOffsetInM
-        )
-
         # Calculate the wavefront signal. Enforce the element outside the mask
         # to be 0.
         den[den == 0] = np.inf
 
         # Calculate the wavefront signal
-        S = num / den / deltaZ
+        S = num / den
 
         # Extend the dimension of signal to the order of 2 for "fft" to use
         padDim = self.getFftDimension()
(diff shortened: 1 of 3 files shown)


teweitsai:
Unless there is any significant improvement of performance, I thought the original implementation fits the paper's algorithm better and easier to understand. In addition, the main variable name in this PR might be misleading in the physics or math. For example, this PR has `dIdz - the gradient image, (dI/dz) = +/- (I1 - I2) / (2 * dz)`. If we do not read the detail, `dIdZ` is more related to a function integration with `dI` and `dz`. But based on the "gradient image", the variable name should be `dI_over_dz` or something similar to avoid any possible misunderstanding.

In addition, if you do want to use the `dI_over_dz`, there are some variable names would need to change to reflect the change of physical amount. For example, the modification of `F`, `S`, `zc_tmp`, etc. in this PR.

Therefore, if this PR is "really" an intension and improvement, please take time to rename all the "possible" affected physical variables and equations in code's comments to reflect the physics correctly to avoid any confusion for future's maintainers. Thanks!

jfcrenshaw:
Thanks for the review! First I'll address the smaller points, before explaining the motivation for this PR.

> If we do not read the detail, `dIdZ` is more related to a function integration with `dI` and `dz`. But based on the "gradient image", the variable name should be `dI_over_dz` or something similar to avoid any possible misunderstanding.

That makes sense. I'd be happy to rename the variable to `dI_over_dz`

> In addition, if you do want to use the `dI_over_dz`, there are some variable names would need to change to reflect the change of physical amount. For example, the modification of `F`, `S`, `zc_tmp`, etc. in this PR.

I don't understand what you are saying here. This change does not have any impact at all on the values, interpretations, units, etc. of any of those variables. Maybe it will make more sense after the explanation below?

---

The point of this PR is to remove code duplication and clarify the logic of the TIE solvers. The TIE relates the central image, `I0`, and the gradient of the image, `dI/dz`, to the wavefront of the system. Our TIE solvers use an image on either side of focus to approximate `I0` and `dI/dz`, and then invert the TIE to solve for the wavefront.

Both "exp" and "fft" use the two defocal images in the exact same way. They approximate `I0 = (I1 + I2) / 2` and `dI/dz = (I1 - I2) / (2 * dz)`. The only way the two solvers differ is that, after you have approximated `I0` and `dI/dz`, they use different strategies for inverting the TIE. 

In the current implementation, the approximation of `dI/dz` from `I1` and `I2` is pretty opaque:
- The simple calculation of `dI/dz` is spread across three different functions, each at varying levels of abstraction
- The "exp" and "fft" solvers calculate `dI/dz` in two different locations, duplicating non-trivial code
- The way in which "exp" and "fft" calculate `dI/dz` is cosmetically different, with different factors appearing in different locations. This might confuse people into thinking the `dI/dz` calculation differs in some way between "exp" and "fft", when in fact, they are identical
- The current implementation creates the intermediate product `dI`, which serves no purpose other than to eventually calculate `dI/dz`

All of the above make the current implementation needlessly confusing, which makes the code hard to understand and maintain. This PR fixes these issues.

teweitsai:
In the `algorithm.py`, the code: `zc_tmp = np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0] / dz` is modified to be `zc_tmp = np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0]` in this PR. I do not think these two `zc_tmp` variables are in the same physical amount. This applies to other variables I pointed out.

Although I could understand you would like to simplify the code, you might need to be very careful for each physical amount. The original implementation tried to be consistent with the equations in papers.

python/lsst/ts/wep/cwfs/algorithm.py
```diff
@@ -1199,14 +1198,9 @@ def _createSignal(self, I1, I2, cliplevel):
             Approximated wavefront signal.
         """
 
-        # Calculate I0 and dI
-        dI, I0 = self._getdIandI0(I1, I2)
-
-        # Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be
-        # -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)
-        # we will ignore the (1/delta z) until later in this method
```
teweitsai, line 1207: Why do you want to remove `Wavefront signal S=-(1/I0)*(dI/dz) is approximated to be -(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)`? Is it wrong in this PR? To maintain the code with scientific calculation, I would prefer to maintain as more calculation details as possible.

jfcrenshaw:
I understand your nervousness about changing the values of the zernikes! Luckily we have unit tests that track the values of zernike predictions. The unit tests confirm these changes make no numerical difference in the zernike estimation. 

The `dz` that I removed here is an example of what I meant when I said "factors appearing in different locations." This is the same factor of `deltaZ` that appears [here](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1227-L1231) in the "fft" algorithm. There is an extra confusion that `dz = 2 * deltaZ`, because the "fft" algorithm has consigned the factor of 2 to [this line](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1209). The new implementation moves the factors of 2 and `dz` to [this line](https://github.com/lsst-ts/ts_wep/blob/312a0eb67b8ad23d375455b73cebdd60ee64ad89/python/lsst/ts/wep/cwfs/algorithm.py#L1266), during the calculation of `dI/dz`.

The only changes that matter numerically in this PR are just moving around factors using the commutativity of multiplication. And these factors are *only* moved between intermediate products that are *never recorded*. Zernike values, signal values, wavefront values, etc. are all left unchanged.

Assuming the reference paper is [Xin et al., Appl. Opt. 54, 9045-9054 (2015)](https://arxiv.org/abs/1506.04839), we are not diverging from the paper at all. Eq. 4 implicitly defines the approximations of `I0` and `dI/dz` that are used in both the "fft" and "exp" algorithms. The only way the new implementation "differs" is by explicitly defining these values, since they are needed by both "exp" and "fft".

python/lsst/ts/wep/cwfs/algorithm.py
jfcrenshaw: The comment `Wavefront signal S=-(1/I0)*(dI/dz)` defines the value we are calculating (the signal `S`), in terms of the values we are calculating it from (`I0` and `dI/dz`).

The rest of that comment `-(1/I0) * dI/(2 * delta z) = (-dI)/(2 * I0) * (1/delta z)` is nothing more than an implicit definition of the approximations we use for `I0` and `dI`. This doesn't belong here because
1. `dI` doesn't even exist in the new implementation
2. The approximations for `I0` and `dI/dz` are *explicitly* defined in the function that calculates them. If we want, we can refer the reader to their definitions there. But it doesn't make sense to include an implicit definition that we are not even directly using at this point in the code.

teweitsai:
1. The unit test is to verify the code's correctness and is unrelated to the truth that a physical amount contains the value and unit. Therefore, it would not be a good reason to let the `zc_tmp` to have a different unit and physical meaning in the new code. It is very significant that the new `zc_tmp` is the original `zc_tmp` times the `dz`, which is not the same compared with the original one. Therefore, you would need to rename it to reflect the reality.

2.  `zc_tmp` is just an example and this PR affects multiple "physical" variables  like this, and you would need to make it right.

3. To have a good variable name is not easy in scientific calculation. Take the `dI/dz` as an example. You could name it as `dI_over_dz` as a straight forward one. However, to make the physical sense, the better name might be `differential_I`, `partial_I`, `gradient_I`, or others to reflect the physics inside (and depend on the dimension). My main point in this PR is to make the affected variables to have meaningful physical names, which does not happen in the current implementation.

jfcrenshaw:
# 1 
> the new zc_tmp is the original zc_tmp times the dz

I want to emphasize again that this statement is **_not_** true. The new `zc_tmp` is identical in every way to the original `zc_tmp`.  The new implementation simply moves the `dz` inside the linear system, which is the exact same thing because `dz` is a constant.

Let me try to explain explicitly with the code and corresponding equations.

### Original implementation

In [line 1168](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1168) of the original implementation, we have this code

```python
np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0]
```

which is solving the equation

$$
d\Omega \sum_a \sum_b (dI)\_{ab}  ~ Z\_{j,ab} 
= d\Omega \sum\_k \beta_k \left[ \sum\_a \sum\_b (I\_0)_{ab} ~ (\nabla Z\_j)\_{ab} ~ (\nabla Z\_k)\_{ab} \right]
$$

where $j,k$ index Zernikes, and $a,b$ index pixels. This yields the temporary values $\beta^k$, which are then divided by `dz` to get `zc_tmp`:

```python
zc_tmp = beta / dz
```

### New implementation

The new implementation brings the code in closer alignment with the equations in the paper by moving the `1 / dz` inside the system of linear equations.

If you divide both sides of the equation above by `dz`, you get

$$
d\Omega \sum_a \sum_b \left( \frac{dI}{dz} \right)\_{ab}  ~ Z\_{j,ab} 
= d\Omega \sum\_k \left( \frac{\beta}{dz} \right)\_k \left[ \sum\_a \sum\_b (I\_0)_{ab} ~ (\nabla Z\_j)\_{ab} ~ (\nabla Z\_k)\_{ab} \right]
$$

Unlike the first equation, this new equation directly matches equations 19-21 from the paper. Now when we solve this system by calling

```python
np.linalg.lstsq(Mij[:, idx][idx], F[idx], rcond=None)[0]
```

we immediately get `beta / dz` which equals `zc_tmp`. Therefore, we are directly solving for `zc_tmp`, and don't need the temporary values $\beta$ which then have to be divided by `dz` after the fact.

# 2 
> this PR affects multiple "physical" variables like this

Again, I will re-emphasize that this statement is _**not**_ true. Hopefully the explanation above makes it clear why no physical variables are changed in the "exp" algorithm. I will now explain in detail why no physical variables are changed in the "fft" algorithm.

The "fft" algorithm uses `I_0` and `dI/dz` to calculate the signal $S$ in [line 1030](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1030) of the original implementation. After this point, everything is calculated in terms of `S`, so all we have to do is show that the value/interpretation of `S` is unchanged.

In the method `_createSignal`, the [original implementation](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1179-L1244) calculates the signal like this

```python
num = -dI  # line 1208
den = 2 * I0  # line 1209
S = num / den / deltaZ  # line 1238
```

The [new implementation](https://github.com/lsst-ts/ts_wep/blob/312a0eb67b8ad23d375455b73cebdd60ee64ad89/python/lsst/ts/wep/cwfs/algorithm.py#L1178-L1230) just regroups some of these terms, like this

```python
num = -dI_over_dz  # line 1202
den = I0  # line 1203
S = num / den  # line 1224
```

where `dI_over_dz = dI / (2 * deltaZ)`. Obviously this regrouping does not change the value or interpretation of `S`.

The only remaining concern is with the thresholding of `den` in [lines 1214-1235](https://github.com/lsst-ts/ts_wep/blob/751d5738eddd1cee7c8906c8ddaf9dd28a01a65f/python/lsst/ts/wep/cwfs/algorithm.py#L1214-L1235). You might worry that since the new `den` is only half the original `den`, that these lines might be impacted. However, these lines only threshold `den` using relative values, not absolute values. Since all pixels in `den` are scaled identically, this scaling will not impact the relative thresholding.


# 3

I have explained why the other "physical variables" are totally unchanged by this PR, and thus their names don't need to be changed. I agree about renaming `dIdz` to something clearer, and will do this today.

teweitsai:
I think the point 1 comes from the changed `F` to be `F = np.tensordot(dIdz, zk, axes=((0, 1), (1, 2))) * dOmega` instead of `F = np.tensordot(dI, zk, axes=((0, 1), (1, 2))) * dOmega`. Doesn't it?

teweitsai:
In addition, it is better to organize the above discussions into the code's comment clearly for the future's follow-up. It is clear that the equations in comments had been modified in this PR.

python/lsst/ts/wep/cwfs/algorithm.py
```diff
@@ -1141,7 +1148,7 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
 
             # Eqn. (19) from Xin et al., Appl. Opt. 54, 9045-9054 (2015).
             # F_j = \int (d_z I) Z_j d_Omega
```
teweitsai: Based on the equation in conversation, this comment of `F_j` may need to be modified. Or do I misunderstand the equation?

jfcrenshaw:
> I think the point 1 comes from the changed `F` to be `F = np.tensordot(dIdz, zk, axes=((0, 1), (1, 2))) * dOmega` instead of `F = np.tensordot(dI, zk, axes=((0, 1), (1, 2))) * dOmega`. Doesn't it?

Yes, that's correct!

Edit:
Ah, I realize this was one of the variables you were worried about changing in your original comment -- my apologies. The values of `F` have indeed changed. I don't think this is a problem, since `F` isn't something we record or interpret. It's just an intermediate product we use to calculate the Zernike coefficients a few lines below.

Since we are changing the value of `F` in the code, we *could* change the name. I think that would add confusion instead of reducing it, since this new `F` matches the definition of $F$ in the paper, while the previous value of `F` *did not* match the definition of $F$ in the paper.

python/lsst/ts/wep/cwfs/algorithm.py
jfcrenshaw: I don't think the comment needs to be changed, since `d_z I` is $\partial I / \partial z$. This PR actually just makes the following line of code, `F = np.tensordot(dIdz, zk, axes=((0, 1), (1, 2))) * dOmega`, match the original comment.

With the previous implementation, the comment probably should have said `# F_j = \int (dI) Z_j d_Omega`, or something similar, where `dI = I2 - I1`, since in the original implementation, the `1/dz` part was reserved until after solving the system of equations.

python/lsst/ts/wep/cwfs/algorithm.py
```diff
@@ -967,7 +969,8 @@ def _solvePoissonEq(self, I1, I2, iOutItr=0):
             Intra- or extra-focal image.
         iOutItr : int, optional
             ith number of outer loop iteration which is important in "fft"
-            algorithm. (the default is 0.)
+            algorithm, and in saving dIdz and I0 to the algorithm history.
```
teweitsai: "saving dIdz and I0" --> "saving dI/dz and I0"

python/lsst/ts/wep/cwfs/algorithm.py
```diff
@@ -1260,23 +1251,31 @@ def _getdIandI0(self, I1, I2):
...
+            * (self._inst.focalLength - self._inst.defocalDisOffsetInM)
+            / self._inst.defocalDisOffsetInM
+        )
+
+        # Calculate the image gradient and mean image
+        partialI = (I1.getImg() - I2.getImg()) / (2 * deltaZ)
         I0 = (I1.getImg() + I2.getImg()) / 2
 
-        # if I2 is extrafocal, flip the sign of dI
+        # if I2 is extrafocal, flip the sign of dIdz
```
teweitsai: "flip the sign of dIdz" --> "flip the sign of dI/dz"

teweitsai (approved):

You may need to fix the flake8 failure on Jenkins.

PR #186, opened 2023-05-30
Tickets/DM-39447

gmegh:
Fixing filter names in testData fits files from "g" to the new "g_6". 

For all the testData/gen3TestRepo fits files, the filter value was updated to comply with new obs_lsst ts8 translator convention. The update was performed with the following snippet:

```
from astropy.io import fits
ass =['R00_SW0', 'R00_SW1', 'R04_SW0', 'R04_SW1', 'R40_SW0', 'R40_SW1', 'R44_SW0', 'R44_SW1']
for sens in ass:
    filename = f'/sdf/home/g/gmegias/aos/ts_wep/tests/testData/gen3TestRepo/LSSTCam/raw/all/raw/20211231/MC_H_20211231_006000/raw_LSSTCam_g_MC_H_20211231_006000_{sens}_LSSTCam_raw_all.fits'
    hdul = fits.open(filename)  # open a FITS file
    hdul[0].header['FILTER'] = 'g_6'
    hdul.writeto(filename, overwrite=True)
```

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-6.3.3:
+
+-------------
+6.3.3
+-------------
+
+* Change filter name in testData/gen3TestRepo camera fits files to comply with new obs_lsst convention.
+
 .. _lsst.ts.wep-6.3.2:
 
 -------------


tests/testData/gen3TestRepo/update
suberlak, line 1: Is this file meant to be there?

suberlak:
Hi! Since all checks have passed I see that the update works. But just for reference, if you could paste here the code that you used to make the updates, it would make it useful to keep track of what happened. Also, please make sure to update the `versionHistory` file, and tag the develop branch after merger. Thanks!

suberlak (approved):
Thanks for the update, made  a few small comments in conversation and the empty file.

jbkalmbach:
> Hi! Since all checks have passed I see that the update works. But just for reference, if you could paste here the code that you used to make the updates, it would make it useful to keep track of what happened. Also, please make sure to update the `versionHistory` file, and tag the develop branch after merger. Thanks!

Tags should be made on the `main` branch after merging the `develop` branch into `main` with a no fast-forward merge. If you haven't done that before let me know and I'll walk you through it.

jbkalmbach (changes requested):
What happens when you run `butler query-datasets tests/testData/gen3TestRepo/`? When I do it the `physical_filter` of all the data is still 'g' and not 'g_6' as I think it should be? Or is that not right?

jbkalmbach (approved):
We can leave the database as is for now since these tests are passing and fighting the butler isn't the point of these tests. Just want us to keep in mind this disparity between the butler database and the actual headers of our fits files in case it comes back to bite us later.

gmegh:
Perfect! Thanks to both, @jbkalmbach and @suberlak . Merging

PR #189, opened 2023-06-08
[tickets/DM-39590] Fix empty catalog output from donutSourceSelectorTask

jbkalmbach:
The output from empty catalogs was causing donutSourceSelectorTask to fail when running the task because it was not outputting a `Struct` with the same format as when there is a catalog with sources. The following changes were made to fix this:

* Make sure output from empty catalogs match that expected from catalogs with sources in donutSourceSelectorTask. 
* Add tests for run method.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,15 @@
 Version History
 ##################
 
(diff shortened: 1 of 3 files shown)


suberlak (approved):
Thanks for the  fix, looks good!

PR #190, opened 2023-06-15
DM-39624: Move fits files to LFS.

jbkalmbach:


jbkalmbach:
This PR moves the files to `git-lfs` but we will still need to edit git history to remove large files from history.

Jenkinsfile
```diff
@@ -11,6 +11,7 @@ pipeline {
           image 'lsstts/develop-env:develop'
           args "--entrypoint=''"
           alwaysPull true
+          label 'Node3_4CPU  || CSC_Conda_Node'
```
teweitsai, line 16: Please help to comment why this special node is needed instead of letting the Jenkins server to assign itself. Thanks

Jenkinsfile
```diff
@@ -46,7 +47,7 @@ pipeline {
         stage('Cloning Repos') {
             steps {
                 dir(env.WORKSPACE + '/ts_wep') {
-                    checkout scm
+                    checkout scmGit(branches: [[name: 'refs/heads/'+env.CHANGE_BRANCH]], extensions: [lfs()], userRemoteConfigs: [[credentialsId: '14e4c262-1fb1-4b73-b395-5fe617420c85', url: 'https://github.com/lsst-ts/ts_wep.git']])
```
teweitsai, line 52: I do not know it is a good idea or not to expose this `credentialsId` in a public repository. Is Rob OK for this? Thanks!

teweitsai (approved):
Looks good. Thanks for this great improvement!

teweitsai:
Please help to update the version history file as well. I did not check the commit: 803c321 because my web browser says the file change is infinity and does not show me the details.

Jenkinsfile
rbovill: @teweitsai The git-lfs tool is only installed on these two nodes.  If the job runs on the other nodes, it will fail.

Jenkinsfile
rbovill: @teweitsai it doesn't really matter either way, I don't think.  It's just an id reference to the credentials, instead of the name and neither is sensitive information (the credentials themselves are sensitive, but not the name/id).

Jenkinsfile
teweitsai: @jbkalmbach , please help to add the comment in `Jenkinsfile` for the details from Rob. Thanks!

PR #197, opened 2023-08-04
Tickets/DM-40273: update comCam opticalModel

suberlak:
Optimize default optical model used for comCam to onAxis.


jbkalmbach (approved):
LGTM

PR #201, opened 2023-09-20
DM-40685: Update ts_pre_commit_config

suberlak:
Adding `ruff` to follow https://tssw-developer.lsst.io/procedures/pre_commit.html#main-content

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,9 +17,9 @@ jobs:
 
       - name: Install
         run: |
-          $CONDA/bin/conda install -c lsstts ts-pre-commit-config=0.4 -y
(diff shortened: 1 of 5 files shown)


jbkalmbach (approved):
If all these parameters are consistent with T&S software elsewhere then this looks good to me. Only comment is that since this doesn't affect user API I think this should be v6.4.12 instead of 6.5.0.

suberlak:
All changes follow https://tssw-developer.lsst.io/procedures/pre_commit.html#main-content

suberlak:
Set `ruff:true`  since from version  `ts-pre-commit-config-0.9.0 `   the  `.ruff.toml` config files don't get overwritten at the github action stage. This means that ruff uses the config file both at USDF (local) and at github (before CI).

PR #202, opened 2023-09-27
DM-40905: fix GenerateDonutFromRefitWcsTask

suberlak:
Add `coord_raErr`, `coord_decErr` fields to a table passed to `AstrometryTask.run` to refit WCS from the direct detect donut catalog, to work with updates from `lsst_distrib` `w2020_38` https://lsst-dm.github.io/lsst_git_changelog/weekly/w_2023_38.html

### DIFF
--- a/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
+++ b/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
@@ -178,6 +178,19 @@ class GenerateDonutFromRefitWcsTask(GenerateDonutCatalogWcsTask):
 
         sourceSchema = afwTable.SourceTable.makeMinimalSchema()
         measBase.SingleFrameMeasurementTask(schema=sourceSchema)  # expand the schema
+        # add coord_raErr,  coord_decErr to the schema
+        sourceSchema.addField(
+            afwTable.Field["F"](
+                name="coord_raErr", doc="position err in ra", units="rad"
+            ),
+        )
+        sourceSchema.addField(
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -201,6 +216,13 @@ def formatDonutCatalog(self, catalog):
             dec = lsst.geom.Angle(catalog["coord_dec"].iloc[i], lsst.geom.radians)
             src.set(sourceDecKey, dec)
 
+            # set default 1% for raErr, decErr
```
jbkalmbach: Can you add a check here for `coord_raErr` and `coord_decErr` in the catalog in case they are already present so we're not overwriting data we have?

jbkalmbach (commented):
Only one comment on what you've added so far, but I think while we're looking at this task, it doesn't currently have a test for this function. Could you add a quick test for this function in the test suite? Just something that checks the columns of the output starting from a small input test dataframe?

python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
suberlak: Added a test that checks for that, and uses the existing `coord_raErr`, `coord_decErr` if they are present, and calculates them as 1% of `coord_ra`, `coord_dec` if these columns are missing.

jbkalmbach (approved):
Thanks, looks good!

PR #203, opened 2023-09-28
Move rotations of donut stamps from calculate step to save donutStamps.

jbkalmbach:


python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -234,38 +224,24 @@ def estimateZernikes(self, donutStampsExtra, donutStampsIntra):
...
-                imageIntra = rotate(
-                    donutIntra.stamp_im.getImage().getArray(), eulerZIntra
-                )
-
+            # Transform image and coordinates from DVCS
+            # (Data Visualization Coordinate System)
+            # to ZCS (Zemax Coordinate System).
             wfEsti.setImg(
-                fieldXYExtra,
+                np.array([-fieldXYExtra[1], fieldXYExtra[0]]),
```
suberlak, line 259: So before `fieldXYExtra` was without a transpose, and this operation does a left-right flip + transpose?  Is it consistent with  left-right flip of the transpose of the donut image below?  Would it be better to move that to `donutExtra.calcFieldXY()` ?

suberlak: (so that neither the image nor the coordinates need any flips at this stage, i.e. we're passing correct image+coords to `calcZernikesTask`, and flipping donuts + coordinates somewhere else - I thought that's the intention of this PR)

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -234,38 +224,24 @@ def estimateZernikes(self, donutStampsExtra, donutStampsIntra):
...
                 DefocalType.Extra,
                 filterLabel=getFilterTypeFromBandLabel(donutExtra.bandpass),
-                image=imageExtra,
+                image=np.fliplr(donutExtra.stamp_im.image.array.T),
                 blendOffsets=blendOffsetsExtra.tolist(),
```
suberlak, line 263: Same as above - we take rotation away in one place, but introduce it here, and i wonder if it makes anything easier...

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -110,13 +109,6 @@ class CalcZernikesTaskConfig(
     instPixelSize = pexConfig.Field(
         doc="Instrument Pixel Size in m", dtype=float, default=10.0e-6
     )
-    transposeImages = pexConfig.Field(
```
suberlak: Perhaps worth leaving some sort of deprecation comment? Otherwise it'll break all other code that was using `calcZernikesTaskConfig.transposeImages` setting - would be good to suggest what's the alternative now.

tests/task/test_calcZernikesTaskCwfs.py
```diff
@@ -158,25 +158,25 @@ def testEstimateCornerZernikes(self):
         ).combinedZernikes
         trueZernCoeffR04 = np.array(
             [
```
suberlak, line 160: So updating the hard-coded test values is a bit worrisome - why are they so much different?  Can you add in the comments a plot of these before and after - and why would merely moving rotations to a different point in the code change the result of Zk fitting?  I thought it was just moving the rotation to a different part of the code, but not changing anything else that would actually change the value of the fit...

suberlak: Looking below I see that the actual test donut stamp images were changed too. Why was that needed?

suberlak (approved):
Some comments, please have  a look, thanks for update!

python/lsst/ts/wep/task/calcZernikesTask.py
teweitsai: I think you might need to have some comments or figures to explain this operation of array values: `np.fliplr(donutExtra.stamp_im.image.array.T)`, which is pretty complex and not straightforward.

The original implementation of rotation of Euler angle makes more physical sense to me.
But it would be fine if you want to do it in another way as long as you explain it clearly.

Thanks!

tests/task/test_calcZernikesTaskCwfs.py
teweitsai: If the test results differ so much compared with the original one, I suspected this PR may not be compatible with PhoSim anymore. Did you check the donut image with `imsim`? If this update is specific to `imsim` and will break the compatibility with PhoSim, I think maybe merge to another `develop-imsim` branch first before the transition is done.

suberlak (approved):

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: I would like to keep the output of `calcFieldXY` in the same coordinate system as the image.

The intention of this PR is to line up the donut stamps with the science sensors not align them to another coordinate system. This way if you save postage stamps for the corner wavefront sensors and the science sensors they are all aligned along the same axes.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: I will add documentation. I would also like to point out that this change in transpose plus lr-flip is in addition to the rotation with the Euler angle.

tests/task/test_calcZernikesTaskCwfs.py
jbkalmbach: I could just rotate the existing test donut stamps so that they are lined up as output with the new donut stamps. I added new test data since the old data was phosim and figured we should have up to date test data with proper shapes but it's not necessary.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Good call. I can add that.

jbkalmbach: For our future reference, info on deprecating configs is found here: https://developer.lsst.io/stack/deprecating-interfaces.html#config-deprecation

tests/task/test_calcZernikesTaskCwfs.py
jbkalmbach: Actually, @teweitsai you are right. Because Phosim doesn't require the left-right flip since it uses a different coordinate system this will be an imsim specific change. What exactly are you suggesting with `develop-imsim` and what would be the process of moving this over to `develop` eventually?

jbkalmbach: On the other hand, we are no longer using phosim so is that really necessary?

teweitsai: Two options:
1. From the current `develop` branch, you could checkout a new `develop-imsim` branch. For the `imsim` related update, `merge` to this `develop-imsim` branch first before the fully transition to `imsim`.
2. Add a note to the `README` and version history to say from this `version`, `PhoSim` is not supported anymore.

To answer you question, it would depend on how long you still need to fully transition to `imsim`. And in this time period, is it possible that someone might need to run all pipeline in `PhoSim` to test or verify something or not. If you still need several weeks or months to finish the transition, I would prefer the option 1 to avoid the lose of continuity.

Note. If you end to have the `develop-imsim` branch, you might need to check the update of `develop` branch and `rebase` to it regularly to make the final `merge` between `develop-imsim` and `develop` branches easier.

jbkalmbach: Sounds good. I think we will be ready with a release version of `ts_imsim` in the next few days so I think it would be good to go with option 2.

jbkalmbach:
Since this will break convergence with https://github.com/lsst-ts/ts_phosim and move us completely to ts_imsim we have conducted studies to show that there is a similar level of convergence between ts_phosim and ts_imsim results. See https://jira.lsstcorp.org/browse/DM-40867 for plots.

PR #204, opened 2023-10-06
Organized util functions in utils directory.

jfcrenshaw:
This is the first PR towards merging the new implementation of the TIE into ts_wep. This PR just centralizes all of the utility functions into a utils module. 

The PR touches a lot of files, because it
(1) moves and organizes the util functions in the utils module
(2) changes lots of imports in ts_wep to `from lsst.ts.wep.utils import ...`
(3) moves the corresponding unit tests around to reflect the restructuring

However, this PR makes no real changes to any code or tests. It merely moves things around for better organization, so that the changes made by the big TIE refactor will touch the fewest number of files possible.


doc/versionHistory.rst
```diff
@@ -5,6 +5,15 @@
 ##################
 Version History
 ##################
+
+.. _lsst.ts.wep-6.4.13:
+
+-------------
+6.4.13
```
jbkalmbach: This should probably be a major version update since it will break backwards compatibility. A little annoying but I think each of your updates may have to increment the major version number for the same reason. It would be better if we had structure to put everything on a release branch but let's not worry about that now.

jbkalmbach (approved):
Ok, looks good.

doc/versionHistory.rst
jfcrenshaw: Ah yeah, good point. I changed the version to 7.0.0

PR #205, opened 2023-10-10
Tickets/DM-41022: fix generateDonutDirectDetect

suberlak:
Adding an empty `blend_centroid_x`, `blend_centroid_y` columns to a donut catalog in case `donutSelector` was not run.


jbkalmbach (approved):
Looks good. Thanks for figuring this out.

PR #207, opened 2023-10-26
Modifications to permit running on MacOS

jmeyers314:


jfcrenshaw:
LGTM.

One question: Are the `stdout` and `stderr` arguments you added to `runProgram` required for Mac compatibility, or are you just adding those because they're useful to have in certain contexts?

jfcrenshaw (dismissed):

jmeyers314:
> Are the stdout and stderr arguments you added to runProgram required for Mac compatibility

Actually, it just occurred to me that this is probably a generic `shlex` compatibility problem, and not just on MacOS.  For example, if you try `subprocess.call(shlex.split("ls &> ls.log"), shell=False)`, instead of piping stderr/stdout to ls.log, ls complains that it can't find the files `&>` and `ls.log`.  I guess redirection doesn't work when handing `subprocess.call` a list instead of a string.

There's at least one line in `ts_imsim` where this matters: https://github.com/lsst-ts/ts_imsim/blob/develop/python/lsst/ts/imsim/closedLoopTask.py#L1191.  I'll add that to this ticket and look around for any others.

So thanks for asking!

jfcrenshaw (approved):
LGTM. You just need to rebase and bump the version patch number once more.

PR #209, opened 2023-10-31
Add local linear WCSs to DonutStamps.

jmeyers314:


jmeyers314:
Actually, this isn't working quite yet with ts_imsim (means I need a better test too).  I apparently can't use a dict as a Stamp.archive_element...

jmeyers314:
Okay, looks like using a bare SkyWCS runs, though it's a little less obvious that the archive_element is actually a wcs in this case...

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -479,6 +481,17 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
                 blendCentroidPositions = np.array([["nan"], ["nan"]], dtype=float).T
 
+            # Get the local linear WCS for the donut stamp
+            wcs = exposure.wcs
+            pt = Point2D(finalDonutX, finalDonutY)
+            spt = wcs.pixelToSky(pt)
+            # Be careful to get the cd matrix from the linearized WCS instead
+            # of the one from the full WCS.
+            lin = wcs.linearizePixelToSky(pt, degrees)
+            cd = lin.getLinear().getMatrix()
```
teweitsai: Please consider to have more meaningful names for `pt`, `spt`, `lin`, and `cd`.

jbkalmbach (approved):
Looks good to me. Thanks for adding this!

jbkalmbach:
Did you have to change the definition of donut stamps in the repo? Is that why it is part of the changes included?

python/lsst/ts/wep/task/cutOutDonutsBase.py
jmeyers314: Sure.  Good idea.

jmeyers314:
> Did you have to change the definition of donut stamps in the repo? Is that why it is part of the changes included?

Do you mean the `use_archive=True` bit?  That's part of the StampsBase: https://github.com/lsst/meas_algorithms/blob/main/python/lsst/meas/algorithms/stamps.py#L342

I'm actually considering changing DonutStamp now though.  In particular, adding something like `.getWCS()` would make discovering the WCS much easier than somehow knowing that `.archive_element` is actually the WCS.

jbkalmbach:
> Do you mean the `use_archive=True` bit? That's part of the StampsBase: https://github.com/lsst/meas_algorithms/blob/main/python/lsst/meas/algorithms/stamps.py#L342

I actually meant including the sqlite database in the changes here.
 
> I'm actually considering changing DonutStamp now though. In particular, adding something like `.getWCS()` would make discovering the WCS much easier than somehow knowing that `.archive_element` is actually the WCS.

Yeah, I agree that that would be a better user experience.

jmeyers314:
> I actually meant including the sqlite database in the changes here.

Ugh. Didn't mean to commit that.  Thanks for catching that.

python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -281,3 +281,15 @@ def makeMasks(self, inst, model, boundaryT, maskScalingFactorLocal):
         self.mask_pupil = afwImage.Mask(
             np.array(self.comp_im.mask_pupil, dtype=np.int32)
         )
+
+    def getLinearWCS(self):
```
jbkalmbach, line 285: Can you add a test for this in test_donutStamp.py? Thanks

jbkalmbach (approved):
Change to add `getLinearWcs` looks good. Please add a test for it and then we are good to merge.

PR #211, opened 2023-12-08
Update default config on GenerateDonutFromRefitWcsTask after updates in meas_astrom.

jbkalmbach:


suberlak (approved):
Looks good, thanks for the quick fix!

PR #213, opened 2024-01-22
Tickets/dm 42572: Add background subtraction to pipeline

jbkalmbach:


python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -269,17 +269,14 @@ def makeMasks(self, inst, model, boundaryT, maskScalingFactorLocal):
...
-        # 0 flag in mask is part of image that is not donut
-        # 1 flag in mask means it is part of the model donut
-        maskDict = {"BKGRD": 0, "DONUT": 1}
-
         # Set masks
-        self.mask_comp = afwImage.Mask(np.array(self.comp_im.mask_comp, dtype=np.int32))
-        self.mask_comp.clearMaskPlaneDict()
-        self.mask_comp.conformMaskPlanes(maskDict)
-        # Will inherit conformed MaskPlaneDict
+        afwImage.Mask.addMaskPlane("DONUT")
```
suberlak, line 273: Is this update related to the background subtraction, or a new way of calculating the mask ?

tests/task/test_cutOutDonutsBase.py
```diff
@@ -247,13 +247,52 @@ def testCalculateFinalCentroid(self):
         self.assertEqual(cornerX, 0)
         self.assertEqual(cornerY, 0)
 
-    def testCutOutStamps(self):
+    def _getExpAndCatalog(self, defocalType):
+        """
+        Helper function to get exposure and donutCatlog for
```
suberlak: `Catlog` typo

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -384,6 +390,10 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         pixelScale = exposure.getWcs().getPixelScale().asArcseconds()
         camType = getCamTypeFromButlerName(cameraName, detectorType)
         bandpass = exposure.filter.bandLabel
+
+        # Run background subtraction
+        self.subtractBackground.run(exposure=exposure).background
+
```
suberlak, line 396: It seems a bit of a black box... I've seen the convergence improvement plots, but can you show donut postage stamps before / after to show that the subtraction is done correctly? Either in the github discussion or the ticket comments.

suberlak (approved):
Thanks for the update!

python/lsst/ts/wep/task/donutStamp.py
jbkalmbach: Yeah, I had to clean up how the DONUT mask plane is added or it created problems with the background subtraction task. The original way created a new mask plane dictionary that had only DONUT and BKGRD but now we just add a DONUT mask plane to the original mask plane dictionary. It's much cleaner now and I think more along the lines of how the code is supposed to be run.

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: stamp_bgsub_compare
Yes I can show. Here is the donut stamp using the old version and the new version. The colorbar shows the change from the subtraction but it is hard to see so here below is a version using Asinh normalization.
stamp_bgsub_compare_asinh

PR #214, opened 2024-01-26
DM-42464: Persistence of Mask Objects in Stamps

jbkalmbach:
Remove mask_comp and mask_pupil from DonutStamp since they don't persist and mask is already contained in MaskedImage stamp.

### DIFF
--- a/python/lsst/ts/wep/task/cutOutDonutsBase.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsBase.py
@@ -26,6 +26,7 @@ __all__ = [
 ]
 
 import lsst.afw.cameraGeom
+import lsst.afw.image as afwImage
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
 import numpy as np
@@ -519,9 +520,10 @@ class CutOutDonutsBaseTask(pipeBase.PipelineTask):
                 bandpass=bandpass,
                 archive_element=linear_wcs,
             )
-            boundaryT = 1
             maskScalingFactorLocal = 1
-            donutStamp.makeMasks(
+            # Just put 1 for boundaryT. Only going to save pupil mask anyway.
+            boundaryT = 1
+            donutStamp.comp_im.makeBlendedMask(
                 inst, self.opticalModel, boundaryT, maskScalingFactorLocal
             )
 
@@ -556,7 +558,17 @@ class CutOutDonutsBaseTask(pipeBase.PipelineTask):
                 donutStamp.stamp_im.image.array, eulerZ
             )
 
-            donutStamp.stamp_im.setMask(donutStamp.mask_comp)
+            # Set masks
+            afwImage.Mask.addMaskPlane("DONUT")
+            donutMaskVal = afwImage.Mask.getPlaneBitMask("DONUT")
+            # Save pupil mask so that other tasks that use the mask can
+            # expand mask borders different amounts to better fit needs.
+            donutStamp.stamp_im.setMask(
+                afwImage.MaskX(
+                    np.array(donutStamp.comp_im.mask_pupil, dtype=np.int32)
+                    * donutMaskVal
+                )
+            )
 
             finalStamps.append(donutStamp)
 
--- a/tests/task/test_donutStamp.py
+++ b/tests/task/test_donutStamp.py
@@ -19,7 +19,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-import os
 import unittest
 
 import lsst.afw.image as afwImage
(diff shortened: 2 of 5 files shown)


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -519,9 +520,10 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
                 bandpass=bandpass,
                 archive_element=linear_wcs,
             )
-            boundaryT = 1
             maskScalingFactorLocal = 1
-            donutStamp.makeMasks(
+            # Just put 1 for boundaryT. Only going to save pupil mask anyway.
+            boundaryT = 1
+            donutStamp.comp_im.makeBlendedMask(
```
suberlak, line 526: So just to clarify:  we're making both `pupil` and `comp` masks here, which are both used before we get to `CalcZernikes` task. And then   `CalcZernikes` task calling `algorithm.py` re-calculates both masks with adjustable `boundaryT` and `maskScalingFactorLocal` parameters that affect the  extent of the `comp` mask  - is that correct?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -556,7 +558,17 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
 
-            donutStamp.stamp_im.setMask(donutStamp.mask_comp)
+            # Set masks
+            afwImage.Mask.addMaskPlane("DONUT")
+            donutMaskVal = afwImage.Mask.getPlaneBitMask("DONUT")
+            # Save pupil mask so that other tasks that use the mask can
+            # expand mask borders different amounts to better fit needs.
+            donutStamp.stamp_im.setMask(
+                afwImage.MaskX(
+                    np.array(donutStamp.comp_im.mask_pupil, dtype=np.int32)
```
suberlak, line 568: before the mask was set as `donutStamp.mask_comp` , so just switching to `pupil` mask would be `donutStamp.mask_pupil`,  but it is being called from `comp_im` because instead of `donutStamp.makeMasks` we're now using `donutStamp.comp_im.makeBlendedMask`?

tests/task/test_donutStamp.py
```diff
@@ -229,63 +225,3 @@ def testCalcFieldXY(self):
                 )
                 self.assertEqual(donutStamp.comp_im.fieldX, np.degrees(trueFieldAngleX))
                 self.assertEqual(donutStamp.comp_im.fieldY, np.degrees(trueFieldAngleY))
-
-    def testMakeMasks(self):
```
suberlak, line 233: With this test removed, is  the `makeBlendedMask` function already tested somewhere ?

suberlak (approved):
Thanks for the update, it's good to clarify which masks are persisted with the `DonutStamp` in butler repository. If these masks are re-calculated in `algorithm.py` (with possibly different values of  `boundaryT` and `maskScalingFactorLocal`, to what extent is the pupil mask used at this stage? Is it mostly for re-centroiding the postage stamp  from the donut catalog coordinates, before the donut stamp is cut out?

jbkalmbach:
> Thanks for the update, it's good to clarify which masks are persisted with the `DonutStamp` in butler repository. If these masks are re-calculated in `algorithm.py` (with possibly different values of `boundaryT` and `maskScalingFactorLocal`, to what extent is the pupil mask used at this stage? Is it mostly for re-centroiding the postage stamp from the donut catalog coordinates, before the donut stamp is cut out?

Actually looking at `CutOutBase` it looks like the centroid uses a template (that is currently the unpadded (pupil) model mask) and gets the mask from the `CompensableImage` again to actually calculate the mask that gets saved here. So in practice, it looks like the mask we save here is never used even though it's effectively the same as the centroiding mask. We save it here just for later reference at the moment, nothing downstream uses this saved mask at the moment.

jbkalmbach:
> Thanks for the update, it's good to clarify which masks are persisted with the `DonutStamp` in butler repository. If these masks are re-calculated in `algorithm.py` (with possibly different values of `boundaryT` and `maskScalingFactorLocal`, to what extent is the pupil mask used at this stage? Is it mostly for re-centroiding the postage stamp from the donut catalog coordinates, before the donut stamp is cut out?

Actually looking at CutOutBase it looks like the centroid uses a template (that is currently the unpadded (pupil) model mask) and gets the mask from the CompensableImage again to actually calculate the mask that gets saved here. The only place this mask gets used is in deblending if we decide to mask out the blended object in the donut stamp but this is not default behavior and only occurs when using blended objects if `multiplyMask` is turned on.

This mask does not get used anywhere downstream in CalcZernikesTask so it is really just saved for later reference and debugging.

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: See my comment on the main conversation about this. But really these masks don't really get used at all normally.

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: Yes and `donutStamp.makeMasks` did exactly `donutStamp.comp_im.makeBlendedMask` as this is doing.

tests/task/test_donutStamp.py
jbkalmbach: Yes, tested in `tests/cwfs/test_compensableImage.py`: https://github.com/lsst-ts/ts_wep/blob/e1c9b0489b2b4705771237b376aabedf6c46d515/tests/cwfs/test_compensableImage.py#L338-L366

PR #215, opened 2024-02-16
Big WEP refactor

jfcrenshaw:
This is the main and final PR associated with the big WEP refactor.

Per discussions with Josh, and Bryce, and Tiago and due to the interlocking and mutually backwards-incompatible nature of this PR, we have decided to submit this as one big PR, rather than spend weeks devising ways to split it into smaller pieces.

Many of the major changes are summarized in the version history file, but I will summarize many of the most important changes here.

- Instrument geometry info is handled by the new `Instrument` class. This class also (optionally) holds the Batoid model associated with the instrument.
- Donut images are held in the new `Image` class. This class is always meant to hold data in the camera coordinate system, with the wavefront sensors de-rotated. I.e. the native global Batoid coordinate system for the telescope.
- All the top-level information in the `DonutStamp` class is now in the native butler coordinate system (i.e. the data visualization coordinate system with rotated wavefront sensors), and all the coordinate system conversions happen in the `DonutStamp._setWepImage()` method. The `DonutStamp` now has a `wep_im` attribute holding an `Image` object that contains all the info we need for wavefront estimation in the correct coordinate system.
- Mapping between the image and pupil planes is handled by the `ImageMapper` class. This class also handles mask generation. There are still 3 optical models: "offAxis", "onAxis", and "paraxial". The "offAxis" model is now generated on-the-fly using the Batoid model, and so you can safely use "offAxis" at any field angle for any telescope that has a Batoid model in its instrument parameter file. Note the new off-axis model is band-dependent.
- Wavefront estimation is still handled by the `WfEstimator` class. It now has far more configurable parameters, including (i) the max Noll index, (ii) whether to begin estimation with zero or the intrinsic zernikes, (iii) whether to return the full OPD or just the wavefront deviation, (iv) whether to return Zernikes starting with Noll index 4 (the previous standard) or starting with 0 (the Galsim convention), and (v) which unit to return Zernikes in.
- All of these parameters are accessible at the task level, including `saveHistory` which will save the algorithm history in the task metadata. Use `lsst.ts.wep.utils.convertMetadataToHistory` convert the json-compatible metadata history back to the native history format.
- There is a general `WfAlgorithm` class that plugs into the `WfEstimator` to estimate Zernikes. The new TIE algorithm is a subclass of `WfAlgorithm`. It contains most of the same config parameters, plus a few additional ones related to centering the images, plus the `maskKwargs` parameter which gives lots of freedom to control the underlying masks used. For example, `maskKwargs=dict(maskBlends=True)` allows you to mask blends during the TIE fitting. All of these parameters are configurable from the task level as well.
- Currently only the "exp" solver is implemented. Andy told me not to worry about the FFT solver for now, especially since the FFT was broken in the old WEP anyway.
- There is a new, more accurate mask model. There is also a fun util for plotting the mask as a function of field angle.

I think this is all of the major changes, but of course feel free to ask questions!

There are two things that have to happen before this can be merged:
- [x] We need to add Batoid to the Jenkins test environment. Wouter van Reeven is helping me with that.
- [ ] I need to figure out why `TestGenerateDonutFromRefitWcsTask.testTaskFit()` fails. 

I will update here as these two things are accomplished, but review of the PR can begin before these are finished as neither should result in any major changes.


policy/instruments/AuxTel.yaml
```diff
@@ -0,0 +1,28 @@
+# LATISS Telescope Parameters
+# lsst.obs.lsst.Latiss
+
+name: AuxTel
+diameter: 1.2 # aperture diameter in meters
+obscuration: 0.3525 # fractional obscuration
```
jmeyers314: I know you write 0.3525 here, but since it was such a pain to get right in the past, maybe elaborate as:

```suggestion
obscuration: 0.3525 # fractional obscuration ( = 0.423 m / 1.2 m)
```

policy/instruments/ComCam.yaml
```diff
@@ -0,0 +1,9 @@
+# LSST ComCam Parameters
+# lsst.obs.lsst.LsstComCam
+
+# Use settings from LsstCam
+imports: policy:instruments/LsstCam.yaml
+
+# Apply these overrides
+name: LsstComCam
+defocalOffset: 1.5225e-3
```
jmeyers314: Why not `1.5e-3` ?

python/lsst/ts/wep/estimation/wfAlgorithm.py
```diff
@@ -0,0 +1,335 @@
...
+            raise ValueError(
+                f"{self.__class__.__name__} requires a pair of "
+                "intrafocal and extrafocal donuts to estimate Zernikes. "
+                "Please provide both I1 and I2."
+            )
+
+        # Validate I1
+        if not isinstance(I1, Image):
+            raise TypeError("I1 must be an Image object.")
+        if len(I1.image.shape) != 2 or not np.allclose(*I1.image.shape):  # type: ignore
```
jmeyers314:
```suggestion
        if I1.image.ndim != 2 or not I1.image.shape[0] == I1.image.shape[1]:  # type: ignore
```

jmeyers314: And similarly I2.

python/lsst/ts/wep/estimation/wfAlgorithm.py
```diff
@@ -0,0 +1,335 @@
...
+            raise TypeError("jmax must be an integer.")
+        if jmax < 4:
+            raise ValueError("jmax must be greater than or equal to 4.")
+
+        # Validate the instrument
+        if not isinstance(instrument, Instrument):
+            raise TypeError("instrument must be an Instrument.")
+
+        # Validate startWithIntrinsic
+        if not isinstance(startWithIntrinsic, bool):
```
jmeyers314, line 142: Probably okay here, but I'll note that I've had issues in the past with checks like this.  To wit:

```
In [15]: isinstance(np.bool_(True), bool)
Out[15]: False
```

python/lsst/ts/wep/estimation/wfAlgorithm.py
```diff
@@ -0,0 +1,335 @@
...
+            raise TypeError("startWithIntrinsic must be a bool.")
+
+        # Validate returnWfDev
+        if not isinstance(returnWfDev, bool):
+            raise TypeError("returnWfDev must be a bool.")
+
+        # Validate units
+        if not isinstance(units, str):
+            raise TypeError("units must be a str.")
+        allowedUnits = ["m", "nm", "um", "arcsecs"]
```
jmeyers314: I like `arcsec` better than `arcsecs` (you wouldn't use `nms`, right?).  Maybe allow both?

python/lsst/ts/wep/estimation/wfAlgorithm.py
```diff
@@ -0,0 +1,335 @@
...
+            Zernike coefficients representing the OPD in meters,
+            for Noll indices >= 4.
+        """
+        ...
+
+    def estimateZk(
+        self,
+        I1: Image,
+        I2: Optional[Image] = None,
+        jmax: int = 22,
```
jmeyers314, line 205: 22 is good until this is merged.  But let's move to 28 in a new PR soon after :)

doc/versionHistory.rst
```diff
@@ -12,7 +12,22 @@ Version History
 9.0.0
 -------------
 
-* Split the ``cwfs`` modules into ``centroid``, ``templates``, and ``estimation``.
+This is a big backwards-incompatible refactor of WEP. The major changes are:
+
+* Split the ``cwfs`` modules into ``centroid``, and ``estimation``.
+* Donut Images are now held by the ``Image`` class. This class is meant to hold information in the de-rotated camera coordinate system.
```
jmeyers314: Is "de-rotated camera coordinate system" the same as "optical coordinate system" [here](https://sitcomtn-003.lsst.io/) ?

policy/instruments/LsstCam.yaml
```diff
@@ -0,0 +1,73 @@
+# LSST Telescope Parameters
+# lsst.obs.lsst.LsstCam
+
+name: LsstCam
+diameter: 8.36 # aperture diameter in meters
+obscuration: 0.612 # fractional obscuration
```
jmeyers314, line 6: I like `0.612`, but I've seen `0.61` before too.  What do `ts_ofc` and `ts_imsim` use?

python/lsst/ts/wep/image.py
```diff
@@ -21,182 +21,342 @@
...
+    """Class to hold a donut image along with metadata.
 
     Parameters
     ----------
-    centroidFindType : enum 'CentroidFindType', optional
-        Algorithm to find the centroid of donut. (the default is
-        CentroidFindType.RandomWalk.)
+    image : np.ndarray
+        The square numpy array containing the donut image.
+    fieldAngle : np.ndarray or tuple or list
```
jmeyers314, line 48: Can we clarify the coordinate system here some more?  Is it OCS, or CCS, for instance?  (In the sense that light from field angle (thx, thy) lands approximately at (scale_factor * thx, scale_factor * thy) in OCS/CCS).

python/lsst/ts/wep/image.py
```diff
@@ -21,182 +21,342 @@
...
+    bandLabel : BandLabel or str, optional
+        Photometric band for the exposure. Can be specified using a
+        BandLabel Enum or the corresponding string. If None, BandLabel.REF
+        is used. The empty string "" also maps to BandLabel.REF.
+        (the default is BandLabel.REF)
+    planeType : PlaneType or str, optional
+        Whether the image is on the image plane or the pupil plane.
+        Can be specified using a PlaneType Enum, or the corresponding string.
+        (the default is PlaneType.Image)
+    blendOffsets : np.ndarray or tuple or list, optional
```
jmeyers314, line 64: In DVCS (since pixels)?

python/lsst/ts/wep/image.py
```diff
@@ -21,182 +21,342 @@
...
+        Raises
+        ------
+        TypeError
+            If the image is not a numpy array
+        ValueError
+            If the array is not square
         """
+        if not isinstance(value, np.ndarray):
+            raise TypeError("image must be a numpy array.")
+        if len(value.shape) != 2 or value.shape[0] != value.shape[1]:
```
jmeyers314:
```suggestion
        if value.ndim != 2 or value.shape[0] != value.shape[1]:
```

python/lsst/ts/wep/image.py
```diff
@@ -21,182 +21,342 @@
...
-                "No 'image' is hold. Use setImg() instead.", category=UserWarning
+        # If None, populate an empty array with the correct shape
+        if value is None:
+            value = np.zeros((0, 2))
+
+        # Convert to float array
+        value = np.atleast_2d(value).astype(float)
+
+        # Check shape
+        if value.shape[1] != 2 or len(value.shape) != 2:
```
jmeyers314:
```suggestion
        if value.shape[1] != 2 or value.ndim != 2:
```

I'll stop marking this one.  I think it's clearer, but feel free to ignore if you disagree.

python/lsst/ts/wep/image.py
```diff
@@ -21,182 +21,342 @@
...
             )
-        else:
-            self.image = image
 
-    def getCenterAndR(self, image=None):
-        """Get the centroid data.
+        self._blendOffsets = value
+
+    @property
+    def mask(self) -> Union[np.ndarray, None]:
```
jmeyers314, line 269: Is `mask` True for source pixels, `maskBlends` True for pixels of blended (interferring) donuts and `maskBackground` True for (unblended) background pixels?  Could possibly clarify (or if you already have, I missed it, sorry).

python/lsst/ts/wep/imageMapper.py
```diff
@@ -0,0 +1,1368 @@
...
+        image: Image,
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """Construct the forward mapping from the pupil to the image plane.
+
+        Parameters
+        ----------
+        uPupil : np.ndarray
+             Normalized x coordinates on the pupil plane
+        vPupil : np.ndarray
+             Normalized y coordinates on the image plane
```
jmeyers314:
```suggestion
             Normalized y coordinates on the pupil plane
```

python/lsst/ts/wep/imageMapper.py
```diff
@@ -0,0 +1,1368 @@
...
+        Returns
+        -------
+        np.ndarray
+            A flattened mask array
+        """
+        # Get the field angle
+        angle = image.fieldAngle
+
+        # Get the angle radius
+        rTheta = np.sqrt(np.sum(np.square(angle)))
```
jmeyers314:
```suggestion
        rTheta = np.hypot(*angle)
```

python/lsst/ts/wep/instrument.py
```diff
@@ -21,399 +21,1028 @@
...
+    refBand : BandLabel or str, optional
+        When getting the wavelength or loading the Batoid model, use this
+        value in place of BandLabel.REF. It should be a BandLabel Enum, or
+        one of the corresponding strings. If set to None, this value defaults
+        to BandLabel.REF. (the default is None)
+    wavelength : float or dict, optional
+        The effective wavelength of the instrument in meters. Can be a float,
+        or a dictionary that corresponds to different bands. The keys in this
+        dictionary are expected to correspond to the strings specified in the
+        BandLabel enum in jf_wep.utils.enums. If set to None, this defaults
```
jmeyers314: ts_wep ?

python/lsst/ts/wep/instrument.py
```diff
@@ -21,399 +21,1028 @@
...
-            return "auxTelZWO"
+    @property
+    def pixelScale(self) -> float:
+        """The pixel scale in arcseconds per pixel."""
+        return np.rad2deg(self.pixelSize / self.focalLength) * 3600
+
+    @property
+    def donutRadius(self) -> float:
+        """The expected donut radius in pixels."""
+        rMeters = self.defocalOffset / np.sqrt(4 * self.focalRatio**2 - 1)
```
jmeyers314, line 428: I would have naively guessed `self.defocalOffset / (2 * self.focalRatio)` (just from drawing triangles...).  What am I missing?

python/lsst/ts/wep/instrument.py
```diff
@@ -21,399 +21,1028 @@
...
+            return np.zeros(jmax + 1)
+
+        # Get the wavelength
+        if len(self.wavelength) > 1:
+            wavelength = self.wavelength[band]
+        else:
+            wavelength = self.wavelength[BandLabel.REF]
+
+        # Get the intrinsic Zernikes in wavelengths
+        zkIntrinsic = batoid.zernike(
```
jmeyers314, line 729: I'd recommend setting a value of `nx=128` or `nx=256` here.  The default (I think 32) is really optimized more for speed than accuracy.  Hopefully the cache here means we can afford a higher-than-default value.

python/lsst/ts/wep/instrument.py
```diff
@@ -21,399 +21,1028 @@
...
+
+        # Get the off-axis model Zernikes in wavelengths
+        zkIntrinsic = batoid.zernikeTA(
+            batoidModel,
+            *np.deg2rad([xAngle, yAngle]),
+            wavelength,
+            jmax=jmax,
+            eps=batoidModel.pupilObscuration,
+            focal_length=self.focalLength,
+            nrad=10,
```
jmeyers314, line 853: Probably okay, but did you check if this `nrad` and `naz` was converged for jmax=78?

python/lsst/ts/wep/paramReader.py
jmeyers314, line 1: So happy this one is disappeared... :)

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -0,0 +1,664 @@
...
+            optical axis. offAxis is recommended when you have a Batoid model
+            and onAxis is recommended when you do not. paraxial is primarily
+            meant for testing (the default is "offAxis")
+
+        Raises
+        ------
+        ValueError
+            If the value is not one of the allowed values
+        """
+        allowedModels = ["onAxis", "offAxis"]
```
jmeyers314: Docs say `paraxial` is valid?

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -0,0 +1,664 @@
...
+
+        Raises
+        ------
+        RuntimeError
+            If the solver is not supported
+        """
+        # Create the ImageMapper for centering and image compensation
+        imageMapper = ImageMapper(instConfig=instrument, opticalModel=self.opticalModel)
+
+        # Re-assign I1/I2 to intra/extra
```
jmeyers314, line 501: I think these lines would be a little easier to read with just a single predicate:
```python
if I1.defocalType == DefocalType.Intra:
   blah blah blah
else:
    blah blah blah
```

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -0,0 +1,664 @@
...
+
+        # Set the caustic and converged flags to False
+        caustic = False
+        converged = False
+
+        # Loop through every iteration in the sequence
+        for i in range(self.maxIter):
+            # Determine the maximum Noll index to compensate
+            # Once the compensation sequence is exhausted, jmaxComp = jmax
+            jmaxComp = next(compSequence, jmax)
```
jmeyers314, line 547: I had no idea `next` could have a default.  Cool :)

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -0,0 +1,664 @@
...
+                extraComp.image /= extraComp.image.sum()  # type: ignore
+
+                # Approximate I0 = I(u, 0) and dI/dz = dI(u, z)/dz at z=0
+                I0 = (intraComp.image + extraComp.image) / 2  # type: ignore
+                dIdz = (intraComp.image - extraComp.image) / (  # type: ignore
+                    2 * instrument.pupilOffset
+                )
+
+                # Estimate the Zernikes
+                zkResid = self._expSolve(I0, dIdz, jmax, instrument)
```
jmeyers314, line 619: Are we giving up on `fft` ?

python/lsst/ts/wep/utils/ioUtils.py
```diff
@@ -58,6 +66,187 @@ def getConfigDir():
...
+        importedConfig = importedConfig | readConfigYaml(path)
+
+    # Apply the overrides to the imported config
+    config = importedConfig | config
+
+    return config
+
+
+def mergeConfigWithFile(configFile: Union[str, None], **kwargs: Any) -> dict:
+    """Marge the passed keyword arguments with the values stored in the file.
```
jmeyers314:
```suggestion
    """Merge the passed keyword arguments with the values stored in the file.
```

python/lsst/ts/wep/utils/maskUtils.py
```diff
@@ -0,0 +1,513 @@
...
+    xc, rc = params
+    resid = (x - xc) ** 2 + y**2 - rc**2
+    return resid[np.isfinite(resid)]
+
+
+def _fitCircle(xPupil, yPupil, xItem, yItem, rPupilOut, rPupilIn, rEdge):
+    """Fit a circle to the item edge on the pupil.
+
+    This function works by taking pairs of points on the pupil and
+    item (e.g. the surface of on of the mirrors), and using these
```
jmeyers314:
```suggestion
    item (e.g. the surface of one of the mirrors), and using these
```

python/lsst/ts/wep/utils/miscUtils.py
```diff
@@ -214,3 +164,216 @@ def extractArray(inArray, dim):
...
+        inside[idx[i]] = _polygonContainsRow(
+            y[idx[i]],
+            xGrid[idx[i]],
+            polyShifted,
+        )
+
+    return inside
+
+
+def polygonContains(
```
jmeyers314, line 328: Just curious, how does this compare (in speed) to the c++ version in `danish`?

python/lsst/ts/wep/utils/zernikeUtils.py
```diff
@@ -20,314 +20,233 @@
...
+
+def createZernikeGradBasis(
+    u: np.ndarray,
+    v: np.ndarray,
+    jmax: int = 28,
+    obscuration: float = 0.61,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Create a basis of Zernike gradient polynomials for Noll indices >= 4."""
+    # Get the Noll coefficient array for the u derivates
+    nollCoeffU = galsim.zernike._noll_coef_array_xy_gradx(jmax, obscuration)
```
jmeyers314: GalSim v2.5 has an API for this now (https://github.com/GalSim-developers/GalSim/blob/releases/2.5/galsim/zernike.py#L916).  Should use that instead of the "private" `_noll_coef_array_xy_gradx` and `_grady`.

python/lsst/ts/wep/utils/zernikeUtils.py
```diff
@@ -20,314 +20,233 @@
...
-
-
-def ZernikeEval(z, x, y):
-    """Calculate the wavefront surface in the basis of Zernike polynomial.
+def zernikeFit(
+    u: np.ndarray,
+    v: np.ndarray,
+    z: np.ndarray,
+    jmax: int = 22,
+    obscuration: float = 0.61,
```
jmeyers314: Was 0.612 a default elsewhere?  Should be consistent.

python/lsst/ts/wep/utils/zernikeUtils.py
```diff
@@ -453,15 +374,64 @@ def ConvertZernikesToPsfWidth(
...
     )
 
     # Convert the Zernike amplitudes from microns to their quadrature
     # contribution to the PSF FWHM
     dFWHM = conversion_factors * zernikes
 
     return dFWHM
+
+
+def getZernikeParity(jmax, axis="x"):
```
jmeyers314, line 410: I think you can do all of this without string processing from the `n, m` indices of a Zernike.  I know I've written things like `if n+m % 2 == 0:` before.

Are we using this somewhere?  I.e., we have a coordinate system adjustment somewhere?

tests/testData/gen3TestRepo/gen3.sqlite3
jmeyers314, line 1: Just making sure this file has legit changed.  I know I've almost accidentally committed trivial non-changes here in the past :)

tests/estimation/test_tie.py
```diff
@@ -0,0 +1,272 @@
...
+    def testBadConvergeTol(self):
+        with self.assertRaises(ValueError):
+            TieAlgorithm(convergeTol=-1)
+
+    def testBadMaskKwargs(self):
+        with self.assertRaises(TypeError):
+            TieAlgorithm(maskKwargs="fake")
+
+    def testAccuracy(self):
+        # Get the test data
```
jmeyers314: I usually like to put tests like this in a loop with the random seed incrementing each iteration.  You're less likely to miss something from bad luck if you test 10 different seeds.

tests/estimation/test_wfEstimator.py
jmeyers314, line 1: I think we may be missing some basic tests (not sure precisely which test_*.py file, but possibly this one).

For instance, we should get consistent results:
  - regardless of the value of `startWithIntrinsic`
  - regardless of the value of `returnWfDev`
  - regardless of the value of `return4Up`
  - ~regardless of the units~ (looks like you have this one).
  - regardless of `jmax` (at least if the higher true zernikes are 0.0).

Maybe they're further down below or I just missed them?  If not, I think these would be great to add.

tests/task/test_calcZernikesTaskCwfs.py
```diff
@@ -209,13 +195,13 @@ def testEstimateCornerZernikes(self):
                 -1.40149246e-04,
                 -4.11223127e-02,
                 -2.42644902e-03,
-                -1.52392233e-01,
+                1.52392233e-01,
```
jmeyers314, line 198: This is due to a coordsys change presumably?  Are we confident in the current signs?

tests/testData/psfGradientsPerZernike/lsstcam.txt
jmeyers314, line 1: Do you know why these changed?

jmeyers314 (commented):
Hey JF,

Looks really good.  I left a lot of comments/questions, but that's just b/c this PR is so massive.  A couple broader questions:

Looks like we're ditching the TIE FFT algorithm?  Or are you planning that for a future PR?  Same thing about the paraxial model; it looks like that's missing in a few places.

I recently learned that ComCam will be vignetted in the corners.  Can we add the ComCam vignetting model parameters to this PR please?  (Sorry if I missed them and they're already there!)

We also might need to support cameras with names `LsstCamSim` and `LsstComCamSim` or something like that.  (DM would like to keep simulated and real cameras separate going forward).  We can revisit that as we encounter it though (which will probably be quickly).

Thanks for all the work on this!  I'm really happy this is in place!

jfcrenshaw:
@jmeyers314 I'm impressed how quickly you did the above review! I'll go through all these comments, but as for your broader questions:

> Looks like we're ditching the TIE FFT algorithm? Or are you planning that for a future PR? Same thing about the paraxial model; it looks like that's missing in a few places.

Andy told me not to worry about FFT for now since it was already broken. It can definitely be added by a future PR, although Andy has told me he doesn't want me to prioritize that.

The paraxial model still exists [here](https://github.com/lsst-ts/ts_wep/blob/869dbd2278bd0398d2e29eff99c4f1901567889f/python/lsst/ts/wep/imageMapper.py#L239-L262). I implemented the paraxial model last though, so I must've missed some places where "paraxial" should be inserted in the docs? Or maybe in allowed values for parameters? I'll go through and look for that.

> I recently learned that ComCam will be vignetted in the corners. Can we add the ComCam vignetting model parameters to this PR please? (Sorry if I missed them and they're already there!)

Sorry, I had not noticed that Batoid has ComCam-specific models. I will switch the ComCam configuration over to those models, and I will generate mask params for ComCam as well.

> We also might need to support cameras with names LsstCamSim and LsstComCamSim or something like that. (DM would like to keep simulated and real cameras separate going forward). We can revisit that as we encounter it though (which will probably be quickly).

That won't be hard to support. We will just need to adjust the util function that maps Camera objects to instrument config files.

policy/instruments/AuxTel.yaml
jfcrenshaw: Done

policy/instruments/ComCam.yaml
jfcrenshaw: I copied the value from [here](https://github.com/lsst-ts/ts_wep/blob/4d89e25a57ecc273dc730f98a5a9dea985d56798/policy/cwfs/instData/comcam/instParamPipeConfig.yaml#L14-L16). Not sure if we should keep this or go back to 1.5mm?

policy/instruments/LsstCam.yaml
jfcrenshaw: It looks like `ts_imsim` uses 0.61, and `ts_ofc` uses inner and outer radii instead of an obscuration value.

python/lsst/ts/wep/image.py
jfcrenshaw: Done

python/lsst/ts/wep/image.py
jfcrenshaw: I agree that ndim is better! I will switch all these.

python/lsst/ts/wep/imageMapper.py
jfcrenshaw: Done

python/lsst/ts/wep/imageMapper.py
jfcrenshaw: Done

python/lsst/ts/wep/instrument.py
jfcrenshaw: Hahah good catch. I was originally developing this in an isolated repo named `jf_wep` ;)

python/lsst/ts/wep/instrument.py
jfcrenshaw: Let $R_p$ be the radius of the pupil on the pupil plane, $R_i$ be the radius of the pupil projected onto the image plane, $\ell$ be the defocal offset, and $f$ be the focal length.

Then we have
```math
R_i = \frac{\ell}{\sqrt{f^2 - R_p^2}} \, R_p = \frac{\ell}{\sqrt{f^2/R_p^2 - 1}} = \frac{\ell}{\sqrt{4 N^2 - 1}},
```
where $N$ is the focal ratio.

In the limit $N \gg 1$, $R_i = \ell / (2 N)$ just like you said. So your naive guess corresponds to slow optical systems. This formula is accurate for fast optical systems. See [SITCOMTN-111](ls.st/sitcomtn-111) for more details.

python/lsst/ts/wep/utils/ioUtils.py
jfcrenshaw: Done

python/lsst/ts/wep/utils/maskUtils.py
jfcrenshaw: Done

python/lsst/ts/wep/utils/zernikeUtils.py
jfcrenshaw: Done

python/lsst/ts/wep/utils/zernikeUtils.py
jfcrenshaw: Yeah I probably should've looked a little closer at Galsim (I just noticed `noll_to_zern`) and done it that way...

No it's never used explicitly by the code. I used it for converting old test data to new test data (although Bryce ended up doing that already on the main branch before this was ever merged), and when I was looking at your Zernike pyramids and trying to figure out potential sign/parity issues

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: Done

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: Hmm interesting. Unless you object, I think I will leave this for now. If this causes problems in the future, we can change it to something more clever.

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: Yeah `arcsec` makes more sense.

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: Agreed!

python/lsst/ts/wep/estimation/tie.py
jfcrenshaw: Thanks!

python/lsst/ts/wep/estimation/tie.py
jfcrenshaw: Done

python/lsst/ts/wep/estimation/tie.py
jfcrenshaw: :)

python/lsst/ts/wep/estimation/tie.py
jfcrenshaw: Yeah, Andy told me to ignore fft since it was already broken anyway. If someone wants to add FFT back in the future, it's pretty trivial to do so in the current structure.

doc/index.rst
```diff
@@ -17,8 +17,7 @@ WEP
 Overview
 ========
```
jbkalmbach, line 18: The page on WEP configuration should be updated as well especially since all our old pipeline runs will break.

python/lsst/ts/wep/utils/ioUtils.py
```diff
@@ -58,6 +66,187 @@ def getConfigDir():
...
+        Absolute path of config file in the policy directory
+    """
+    # Remove policy: prefix if present
+    path = path.removeprefix("policy:")
+
+    # Return the absolute policy path
+    return os.path.join(getConfigDir(), path)
+
+
+def readConfigYaml(path: str, recurseImports: bool = True) -> dict:
```
jbkalmbach, line 91: Easier to just use `pyyaml-include` package? https://pypi.org/project/pyyaml-include/

policy/instruments/LsstCam.yaml
```diff
@@ -0,0 +1,73 @@
...
+focalLength: 10.312 # effective focal length in meters
+defocalOffset: 1.5e-3 # defocal offset in meters
+pixelSize: 10.0e-6 # pixel size in meters
+refBand: r # band to use for BandLabel.REF
+wavelength: # effective wavelength per band, in meters
```
jbkalmbach: Did you calculate these yourself from the latest measured throughputs? Would be good to have a source/ throughputs version on this.

python/lsst/ts/wep/estimation/wfAlgorithm.py
```diff
@@ -0,0 +1,335 @@
...
+            attribute. If True, then self.history contains information
+            about the most recent time the algorithm was run.
+
+        Returns
+        -------
+        np.ndarray
+            Zernike coefficients representing the OPD in meters,
+            for Noll indices >= 4.
+        """
+        ...
```
jbkalmbach, line 199: Use `NotImplementedError` just to be more helpful to anybody that comes after?

python/lsst/ts/wep/estimation/wfAlgorithm.py
jmeyers314: It's probably fine, just want to spread awareness in case somehow it turns out to be not fine.

jbkalmbach (commented):
I focused on reviewing the gen3 structure since I'm more familiar with that. It looks good but you need tests for your `testEstimateZernikes` tasks.

policy/instruments/LsstCam.yaml
jfcrenshaw: I think these came from the bandpasses included in Galsim (which I think are pretty old). Good idea: I will try and get the new Ag-Ag-Ag throughputs and compute new wavelengths

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: That's what the `@abstractmethod` decorator is for, but I can add this if you think it's confusing as currently written

python/lsst/ts/wep/utils/ioUtils.py
jfcrenshaw: oh cool, I didn't know about this! I will look into using it

doc/index.rst
jfcrenshaw: Do you mean [this page](https://github.com/lsst-ts/ts_wep/blob/develop/doc/configuration/configuration.rst)? Everything on there is still correct

python/lsst/ts/wep/estimation/wfAlgorithm.py
jbkalmbach: Up to you. I've seen DM code with both (see meas_algorithms/stamp.py) and it was a little clearer to me but I guess in this case it doesn't make too much difference.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -516,60 +426,17 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
-                afwImage.MaskX(
-                    np.array(donutStamp.comp_im.mask_pupil, dtype=np.int32)
-                    * donutMaskVal
-                )
-            )
+            donutStamp.makeMask(instrument, self.opticalModel)
 
+            if (self.multiplyMask is True) and blendExists:
+                stampMask = donutStamp.stamp_im.mask
+                sourceMask = stampMask.array == stampMask.getMaskPlane("DONUT")
```
jbkalmbach: This should be a check on whether the bit is set or not, not a test for equality. We should have a test that checks the multiply mask because this should not give you anything. `getMaskPlane` gives you the bit number (11), `getPlaneBitMask` gives you the integer value (2048) which is what is stored in the array and will work if we only expect one bit to be set for the donut masks but will not work in general. So, if your donut masks never have more than one bit set a test for equality with `getPlaneBitMask` will actually work but not in the general case.

jbkalmbach: Might be helpful: https://stackoverflow.com/questions/147713/how-do-i-manipulate-bits-in-python?rq=3

python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -208,35 +206,174 @@ def calcFieldXY(self):
...
+        # when reassigning values below
+        # In this array:
+        # -1 = background
+        # -2 = unblended regions of source donut
+        # -3 = unblended regions of other donuts
+        # -4 = blended regions of source & other donuts
+        stampMask = -1 - stampMask
+
+        # Create the mask planes
+        afwImage.Mask.addMaskPlane("BKGRD")
```
jbkalmbach: We should probably use some other term here or at least check if DM uses a BKGRD mask plane when doing background subtraction. Is there a need to mark the background pixels in the stamp mask anyway? Couldn't we just not mark these bits and leave them 0 in the integer array?

policy/instruments/ComCam.yaml
jmeyers314: Oh interesting.  My naive guess is that Bo probably found that he could use the LSSTCam distortion model with ComCam as long as he made the defocal distance 1.5225e-3.  Total guess.

If we're getting the distortion coefficients on the fly though with the actual ComCam model, then I think it'd make more sense to put this back at 1.5e-3.

policy/instruments/LsstCam.yaml
jmeyers314: Okay.  Can you make a new jira ticket to standardize on one of these?  Might as well standardize batoid/danish/batoid_rubin too, which I think currently use a mix unfortunately.

python/lsst/ts/wep/task/donutStamp.py
jmeyers314: Yeah, I'm confused.  Aren't all pixels affected by background?

python/lsst/ts/wep/utils/zernikeUtils.py
jfcrenshaw: Done

python/lsst/ts/wep/utils/zernikeUtils.py
jfcrenshaw: Okay this now uses `noll_to_zern`.

policy/instruments/LsstCam.yaml
jfcrenshaw: I updated these to the v1.9 throughputs, and added a little more documentation about the version number + what is included in the calculation.

python/lsst/ts/wep/image.py
jfcrenshaw: These are in the global CCS. See the new note on coordinate systems at the top of the docstring.

python/lsst/ts/wep/image.py
jfcrenshaw: The coordinate system is the global CCS. See the new note on coordinate systems at the top of the docstring.

doc/versionHistory.rst
jfcrenshaw: I changed this to the "global camera coordinate system" and added more details about this to the top of the `Image` docstring.

python/lsst/ts/wep/image.py
jfcrenshaw: Let me know if this is clear enough or if I should expand on this

python/lsst/ts/wep/image.py
jfcrenshaw: `mask` is 1 for source pixels, 0 for other pixels. `maskBlends` is 1 for blend pixels, 0 for others. Etc. 

I added a note to all the docstrings about this.

python/lsst/ts/wep/instrument.py
jfcrenshaw: Depends on how we want to define "converged". Some of the Zernikes have big fractional changes as you go from `nrad=10` to `nrad=500` (assuming `naz = 6 * nrad`). For example, `Z27` and `Z28` change by 100%! But they're small to begin with.

If we look at image position residuals, with `nrad = 10`:
```python
mean resid: 1.790e-07 m
max resid: 4.271e-07 m
``` 
whereas with `nrad=500`
```python
mean resid: 1.759e-07 m
max resid: 4.071e-07 m
```
The max resid only changes by ~20 nm, and the mean changes by ~4 nm. I think these are negligible considering the pixel size of 10 microns.

Is that a good metric?

python/lsst/ts/wep/instrument.py
jfcrenshaw: Picking a random field point (`(1.03, 0.71)`) and increasing from `nx=32` to `nx=256` only changes the zernikes at the < 2 nm level. So I think `nx=32` is probably fine? I can increase it though if you think this is important.

python/lsst/ts/wep/utils/miscUtils.py
jfcrenshaw: The C++ version is roughly 10x faster.

C++: ~50 microseconds
Numba: ~500 microseconds

As you can see, I don't think either is the limiting factor. Although the numba JIT compilation slows down importing `ts_wep` a little bit.  Is the C++ version just compiled once?

tests/estimation/test_tie.py
jfcrenshaw: Done

tests/estimation/test_wfEstimator.py
jfcrenshaw: Done

tests/task/test_calcZernikesTaskCwfs.py
jfcrenshaw: I'm fairly confident with my current signs. If you look at the PR that Bryce made to remove the Zemax coordinate system, he applied the x-parity flip to the first set of Zernikes but for the second set of Zernikes (those highlighted here) the transformation he applied is _not_ the x-parity flip. I _think_ this was probably a mistake?

The sign flips here correspond to the x-parity flip of the original Zernikes before Bryce's PR from last month.

Maybe @jbkalmbach can comment as to whether the difference was intended or just an accident?

tests/testData/gen3TestRepo/gen3.sqlite3
jfcrenshaw: Yes it appears this was an accident (because all the tests still pass with the original version). I have reverted to the original version.

python/lsst/ts/wep/task/donutStamp.py
jfcrenshaw: Okay I've left the background as zero

python/lsst/ts/wep/utils/ioUtils.py
jfcrenshaw: I'm actually going to leave this for now as it works and is simple. If we want to upgrade the infrastructure to use `pyyaml-include` we can do that on a future PR.

python/lsst/ts/wep/image.py
jmeyers314: Looks good.  Thanks

python/lsst/ts/wep/instrument.py
jmeyers314: Can you check the timing?  I'd put it at the largest power of 2 that doesn't noticeably affect the algorithm speed.

python/lsst/ts/wep/instrument.py
jmeyers314: Yep, that makes sense.  Thanks for checking.  Maybe add a comment reporting the above?

python/lsst/ts/wep/utils/miscUtils.py
jmeyers314: Yeah, the c++ is compiled ahead-of-time so only GitHub actions pay that price :).  Not critical, but maybe we can switch this to the danish version on the danish PR.

python/lsst/ts/wep/instrument.py
jfcrenshaw: I added a note to the docstring on the method that calculates zernikeTA

tests/task/test_calcZernikesTaskCwfs.py
jbkalmbach: It looks like an accident but then the question is why did the test pass? Maybe we should tighten allowed error bounds?

python/lsst/ts/wep/instrument.py
jfcrenshaw: The current TIE takes about 800ms. `nx=128` adds about 24ms, and `nx=256` adds about 102ms. Since the latter is a greater than 10% increase, I have set `nx=128`.

python/lsst/ts/wep/estimation/wfAlgorithm.py
jfcrenshaw: I think because testing a limit on the sum will inherently be far more sensitive to the larger magnitude Zernikes. The changes in sign here are for indices that are at least an order of magnitude smaller than the larger coefficients.

policy/instruments/LsstCam.yaml
jfcrenshaw: https://jira.lsstcorp.org/browse/DM-43068

tests/task/test_calcZernikesTaskCwfs.py
jfcrenshaw: I think it's because we test the sum of the errors. The coefficients you accidentally flipped are all at least an order of magnitude smaller than the larger Zernike coefficients, so these are dwarfed by the larger Zernikes. I think catching these details would require fine-tuning the error limit, or instead doing a per-Zernike check.

jfcrenshaw: I think we should leave this for now, and we can discuss in the AOS meeting if we want to curate some new data with some specific metrics to do the tests.

jfcrenshaw:
Rebased on Bryce's WCS fix in develop branch

PR #219, opened 2024-03-01
tickets/DM-43127: Direct comparison of ImageMapper to Batoid raytracing.

jfcrenshaw:


tests/test_imageMapper.py
```diff
@@ -478,3 +478,124 @@ def testGetProjectionSize(self):
...
+                dirCos = batoid.utils.fieldToDirCos(*np.deg2rad(angle))
+                rays = batoid.RayVector.asPolar(
+                    optic=optic,
+                    wavelength=inst.wavelength[band],
+                    dirCos=dirCos,
+                    nrad=5,
+                    naz=30,
+                )
+
+                # Get normalized pupil coordinates
```
jmeyers314, line 543: This is probably okay given the symmetry in `asPolar`, but I think a more precise way to get the entrance pupil coordinates is to actually intersect that surface.  I actually found a bug with this in the danish test suite recently, but take a look at https://github.com/jmeyers314/danish/blob/focal_plane_hits/tests/test_factory.py#L553-L555 for an accurate way to do this.

jmeyers314 (approved):
LGTM.

Just FYI though, I spent some time this weekend refining some similar tests in danish (see the `focal_plane_hits` branch).  There when modeling the x/y shifts from the gradient of the ZernikeTA output, I got about ~6%-of-a-pixel residuals.  If I model the x and y offsets independently and not as the gradient of a single scalar fn though, I get residuals smaller than 0.5% of a pixel or so.  (https://github.com/jmeyers314/danish/blob/focal_plane_hits/tests/test_factory.py#L553-L555 and https://github.com/jmeyers314/danish/blob/focal_plane_hits/tests/test_factory.py#L358-L363).

tests/test_imageMapper.py
jfcrenshaw: Pupil coordinates now use the stop intersection as you described.

PR #220, opened 2024-03-12
Add LSSTComCamSim as allowed camera type.

jbkalmbach:


.github/workflows/lint.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
 
       - name: Install
         run: |
+          $CONDA/bin/conda install python=3.11 -y
```
suberlak, line 20: if you rebase against develop I think this won't be needed

suberlak (approved):
Do we have any files that would have that instrument name in the header to test that they can be processed ?

.github/workflows/lint.yaml
jbkalmbach: I'll keep`-y` in though to avoid any issues with needing confirmation in github actions.

jbkalmbach:
> Do we have any files that would have that instrument name in the header to test that they can be processed ?

Demonstration that the closed loop creates `LSSTComCamSim` headers and they are ingested in the butler repository as such is found in the JIRA issue: https://rubinobs.atlassian.net/browse/DM-43221

Also shows that the closed loop successfully converges with `LSSTComCamSim` images.

PR #221, opened 2024-03-13
Tickets/DM-43293: fix latiss tests

suberlak:
Fixing latiss tests by switching from `os.getlogin` to `getpass` (the former didn't work on RSP), and updating Zk values. README is also updated to point out that `batoid` > 0.6.0 is required to work with `batoidOffsetValue`

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,18 @@
 Version History
 ##################
 
+-------------
+9.1.1
+-------------
+
+.. _lsst.ts.wep-9.1.1:
+
+* Fix latiss tests by using getpass, and updating Zk values
+
+-------------
+9.1.0
+-------------
+
 .. _lsst.ts.wep-9.1.0:
 
 * Added ``jmin`` arguments to Zernike utility functions.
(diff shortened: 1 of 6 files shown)


tests/task/test_calcZernikesTaskLatiss.py
```diff
@@ -147,28 +148,28 @@ def testEstimateZernikesRegression(self):
...
+                0.0015623,
+                0.01403694,
+                -0.00709998,
+                -0.03391995,
+                -0.02250289,
+                0.0147995,
             ]
         )
-
+        print(zernCoeff.outputZernikesRaw[0])
```
jbkalmbach: This should be removed.

jbkalmbach (commented):
The description indicates there should be some README changes as well. Are those meant to be here? Let me know and I'll finish off the review.

doc/versionHistory.rst
```diff
@@ -6,6 +6,18 @@
 Version History
 ##################
 
+-------------
+9.2.0
```
jbkalmbach: This update doesn't add any new functionality and should only be 9.1.1. See semver.org for guidelines.

jbkalmbach (approved):
Looks good.

PR #222, opened 2024-03-14
Tickets/DM-42228: add donut quality check

suberlak:


doc/versionHistory.rst
```diff
@@ -14,6 +22,8 @@ Version History
 
 * Added unit test directly comparing ``ImageMapper`` optical models to Batoid raytracing.
   
+=======
```
jbkalmbach: This should be removed, possibly this is left over from a merge conflict?

python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -153,12 +153,23 @@ def run(
             units=self.config.units,
             saveHistory=self.config.saveHistory,
         )
+        # Exclude stamps that are not effective
```
jbkalmbach: Can you create a config and make this an optional step? It would be easier for debugging things in the future if we can easily turn this on and off when running the pipeline.

python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -153,12 +153,23 @@ def run(
...
+        donutStampsIntraSelect = np.array(donutStampsIntra)[
+            effectiveCombined.astype(bool)
+        ]
 
         # Loop over donut stamps and estimate Zernikes
         zkList = []
         histories = dict()
         for i, (donutExtra, donutIntra) in enumerate(
-            zip(donutStampsExtra, donutStampsIntra)
+            zip(donutStampsExtraSelect, donutStampsIntraSelect)
```
jbkalmbach: How do we keep track of which Zernike estimates go with which donut in the end? I think we discussed adding in flags to the `CalcZernikesTask` to identify?

python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -174,6 +175,14 @@ def factory(cls, stamp_im, metadata, index, archive_element=None):
                 if metadata.get("BANDPASS") is not None
                 else ""
             ),
+            # "EFFECTIVE" is a measure of entropy-based information content
```
jbkalmbach: Can you add something here to indicate this is a signal-to-noise effectiveness measure? Otherwise it's pretty vague.

python/lsst/ts/wep/task/donutStamps.py
```diff
@@ -179,6 +181,17 @@ def getBandpasses(self):
         """
         return [stamp.bandpass for stamp in self]
 
+    def getEffectives(self):
+        """
+        Get the effectiveness measure for each stamp.
```
jbkalmbach: Also can you add something indicating this is related to SNR?

python/lsst/ts/wep/donutImageCheck.py
```diff
@@ -64,9 +70,15 @@ def isEffDonut(self, donutImg):
         # Square the distribution to magnify the difference in entropy
         imgEntropy = entropy(hist**2)
         if (imgEntropy < self.entroThres) and (imgEntropy != 0):
-            return True
+            eff = True
+        else:
+            eff = False
+
+        # Return the actual entropy value if needed
+        if self.returnEntro:
```
jbkalmbach, line 78: Can you add a new test to check this works correctly?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -86,6 +100,31 @@ class CalcZernikesTaskConfig(
...
+        default=False,
+        doc="Whether to use signal to noise ratio in deciding to use the donut.",
+    )
+
+    minSignalToNoise = pexConfig.Field(
+        dtype=float,
+        default=5000,
+        doc=str(
+            "The S/N threshold to use (keep donuts only above the threshold)."
+            + "The default S/N = 5000 corresponds to the default entropy threshold (3.5)"
```
jbkalmbach: I don't understand why S/N = 5000 is necessarily equivalent to an entropy of 3.5. Where does this come from?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -86,6 +100,31 @@ class CalcZernikesTaskConfig(
...
     )
+    selectWithEntropy = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Whether to use entropy in deciding to use the donut.",
+    )
+
+    selectWithSignalToNoise = pexConfig.Field(
+        dtype=bool,
+        default=False,
```
jbkalmbach: If both `selectWithSingalToNoise` and `selectWithEntropy` are both `True` what is the functionality when it passes only one test? Is it logical and or does one take priority? Please add some more documentation explaining what happens in that situation as well as adding tests to verify that the expected behavior is what you see.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -120,13 +165,75 @@ def run(
             return pipeBase.Struct(
                 outputZernikesRaw=np.full(19, np.nan),
                 outputZernikesAvg=np.full(19, np.nan),
+                donutsExtraQuality=[],
```
jbkalmbach: I think these should be empty dataframes. You should add a test that checks the pipeline runs with empty content because I think this should fail since it doesn't match the output described in the `connections` class.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -120,13 +165,75 @@ def run(
...
+                donutsIntraQuality=[],
             )
+        # Which donuts to use for Zernike estimation
```
jbkalmbach: This logic (down to line 226) should be moved to its own function. This needs tests as well and moving it to its own function will make it simpler to test.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -120,13 +165,75 @@ def run(
...
+        snExtra = np.zeros(len(donutStampsIntra))
+        snIntra = np.zeros(len(donutStampsIntra))
+        if self.selectWithSignalToNoise:
+            snExtra = np.asarray(donutStampsExtra.metadata.getArray("SN_VAR"))
+            snSelectExtra = self.minSignalToNoise < snExtra
+
+            snIntra = np.asarray(donutStampsIntra.metadata.getArray("SN_VAR"))
+            snSelectIntra = self.minSignalToNoise < snIntra
+
+        # AND condition : if both selectWithEntropy
```
jbkalmbach: Ah okay. I see this answers my question in the comments above. Can you just add this to the docstrings for the configs too? Still add a test to verify behavior.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -124,6 +127,12 @@ class CutOutDonutsBaseTaskConfig(
         target=lsst.meas.algorithms.SubtractBackgroundTask,
         doc="Task to perform background subtraction.",
     )
+    bkgDilationIter = pexConfig.Field(
+        doc="How many iterations of binary dilation to run on the image mask "
+        + "to calculate the background variance (The default is 10).",
+        dtype=int,
+        default=10,
+    )
```
jbkalmbach, line 149: Should we have a way to check that there is still some background in the stamp left? Or some max number of iterations based upon the mask and stamp size? Is that a worry?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -261,7 +275,88 @@ def calculateFinalCentroid(self, exposure, template, xCent, yCent):
...
+
+    def calculateSN(self, stamp):
+        """
+        Calculate signal-to-noise ratio.
+
+        Parameters
+        ----------
+        stamp : lsst.ts.wep.task.donutStamp
+            A stamp containing donut image.
+
```
jbkalmbach, line 316: Remove extra empty line

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -261,7 +275,88 @@ def calculateFinalCentroid(self, exposure, template, xCent, yCent):
...
+        dict
+             A dictionary of calculated quantities
+        """
+
+        stamp.makeMask(self.instConfigFile, self.opticalModel)
+        image = stamp.stamp_im.image.array
+        variance = stamp.stamp_im.variance.array
+
+        # The following are example donut mask values:
+        # maskPlaneDict={'BAD': 0, 'BLEND': 10, 'CR': 3, 'DETECTED': 5,
```
jbkalmbach, line 328: Just to be clear. None of these values will show up in the mask when using `makeMask` except 9 and 10 right? A comment to clarify would be good here and help make the values you calculate in the next paragraph make sense.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -261,7 +275,88 @@ def calculateFinalCentroid(self, exposure, template, xCent, yCent):
...
+        # The mean image value  in the background region
+        background_image_mean = np.mean(image[bkgnd_mask])
+
+        # Total noise based on the variance of the image background
+        ttl_noise_bkgnd_variance = np.sqrt(background_image_variance * n_px_mask)
+
+        # Noise based on the sum of variance plane pixels inside the donut mask
+        ttl_noise_donut_variance = np.sqrt(variance[donut_mask].sum())
+        sn = 0
+        # Avoid zero division in case variance plane doesn't exist
```
jbkalmbach, line 387: Does this mean we throw away all donuts in an image where the variance plane doesn't exist? In what cases do we expect the variance plane to have errors like that?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -329,6 +427,21 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         xCornerList = list()
         yCornerList = list()
 
+        # Centroid shift
+        drCentroidList = list()
```
jbkalmbach, line 488: Change to separate dx, dy as mentioned in today's AOS meeting.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -435,14 +566,65 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         stampsMetadata["DFC_DIST"] = np.array(
             [instrument.defocalOffset * 1e3] * catalogLength
         )
+        # Save the donut flux as magnitude
```
jbkalmbach, line 653: What happens when you have this branch set up and you try to load butler datasets with old donut stamps? Do you run into errors?

python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
```diff
@@ -159,32 +161,25 @@ def run(
         camera: lsst.afw.cameraGeom.Camera,
     ) -> pipeBase.Struct:
         cameraName = camera.getName()
+        # The order of donut catalogs is the same
```
jbkalmbach: I think these changes should be rolled back. When running interactively a user might not know that they should put in extra, intra in order. Maybe what we should do is make an issue to find a way to force the `donutCatalogs` to have the same test as the exposures?

jbkalmbach: Could be as simple as including the detector name in our donut catalogs.

tests/task/test_calcZernikesTieTaskScienceSensor.py
```diff
@@ -124,6 +140,33 @@ def testEstimateZernikes(self):
...
+        # check that all desired quantities are included
+        colnames = list(struct.donutsIntraQuality.columns)
+        desired_colnames = [
+            "SN",
+            "ENTROPY",
+            "ENTROPY_SELECT",
+            "SN_SELECT",
+            "FINAL_SELECT",
+        ]
+        np.testing.assert_array_equal(np.sort(colnames), np.sort(desired_colnames))
```
jbkalmbach, line 159: As mentioned above we should test the actual logic inside the selection code.

tests/task/test_cutOutDonutsBase.py
```diff
@@ -268,3 +268,53 @@ def testCutOutStamps(self):
...
+
+        # test that the effectiveness contains only binary values
+        unique_values = set(np.unique(donutStamps.metadata.getArray("EFFECTIVE")))
+        allowed_values = set([0, 1])
+        self.assertTrue(allowed_values.issuperset(unique_values))
+
+        # test the calculation of SN
+        sn_values = [25.88952737640593, 39.90181466115093, 33.6410205949]
+        sn_calculated = donutStamps.metadata.getArray("SN")
+        np.testing.assert_array_almost_equal(sn_values, sn_calculated, decimal=4)
```
jbkalmbach, line 424: Can you add a test that runs the pipeline and verifies all the metadata gets stored in the butler and when you pull the donut stamp from the repository that it still has all the metadata you added? I wonder if because the `DonutStamps` class doesn't know about these if they get loaded correctly.

jbkalmbach (changes requested):
I don't think it needs any really extensive changes. Marking as request changes because there's a lot more tests I'd like to see added and also the existing tests don't seem to be passing on Jenkins.

python/lsst/ts/wep/donutImageCheck.py
suberlak: test added

python/lsst/ts/wep/task/calcZernikesTask.py
suberlak: added a test for that, and changed list to empty DF.

python/lsst/ts/wep/task/calcZernikesTask.py
suberlak: Made a new function and added tests for it

python/lsst/ts/wep/task/cutOutDonutsBase.py
suberlak: Added checking for that

python/lsst/ts/wep/task/cutOutDonutsBase.py
suberlak: They actually do - I've seen them in various variations, so eg. 1024 and 2048 are not the only values.

python/lsst/ts/wep/task/cutOutDonutsBase.py
suberlak: Clarified that we'll use the background variance estimate for SN in that case, added a log message.

python/lsst/ts/wep/task/cutOutDonutsBase.py
suberlak: It would work fine-  all quantities are calculated here, and those that are not (variance plane) can be replaced with background image variance. I added necessary legacy behavior for donutCatalog that do not have 'detector' column in cutOutDonutsCwfs.

python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
suberlak: Added detector name to donut catalog so that we test for order, as well as legacy behavior.

tests/task/test_calcZernikesTieTaskScienceSensor.py
suberlak: added a test for `selectDonuts` function

tests/task/test_cutOutDonutsBase.py
suberlak: Added that to test for `cutOutDonutsScienceSensorTask`, since that's where the pipeline is run.

policy/snLimitStar.yaml
```diff
@@ -0,0 +1,15 @@
+---
+
+# Limits of available minimum SN for star selection
+
+# The values for these are from simulation of isolated stellar grid
+# using both direct detect and WCS-based reference catalog fluxes 
+# See
```
jbkalmbach: It looks like the documentation here is incomplete.

tests/task/test_cutOutDonutsScienceSensorTask.py
```diff
@@ -261,6 +262,61 @@ def testTaskRun(self):
                 donutStamp.stamp_im, cutOutStamp.stamp_im
             )
 
+        # Check that the  new metadata is stored in butler
```
jbkalmbach: Extra space between "the" and "new"

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -120,13 +177,100 @@ def run(
...
+        )
         zkCoeffCombined = self.combineZernikes.run(zkCoeffRaw.zernikes)
 
         return pipeBase.Struct(
             outputZernikesAvg=np.array(zkCoeffCombined.combinedZernikes),
             outputZernikesRaw=np.array(zkCoeffRaw.zernikes),
+            donutsExtraQuality=donutsExtraQuality,
+            donutsIntraQuality=donutsIntraQuality,
         )
+
```
jbkalmbach: What do you think about moving all the selectDonuts code into a subtask that can be loaded like the `combineZernikes` and `estimateZernikes` subtasks? I think it would make things a lot cleaner and be easier to turn on and off if we want to do some debugging. We do the same for selecting donuts when making the original catalogs so I think it makes sense to do it here too.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -446,8 +600,18 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
+
+            # Store entropy-based measure of donut quality
+            eff, entro = donutCheck.isEffDonut(donutStamp.stamp_im.image.array)
+            isEffective.append(eff)
+            stampsEntropy.append(entro)
+
             finalStamps.append(donutStamp)
 
+        # Calculate the difference between original centroid and final centroid
+
```
jbkalmbach: Remove empty line after comment

jbkalmbach (commented):
Sorry, I know you've done a lot of coding already for this but I think we should split off the selection step in `CalcZernikes` into a subtask.

python/lsst/ts/wep/task/calcZernikesTask.py
jbkalmbach: Like we do here: https://github.com/lsst-ts/ts_wep/blob/8ac739e3d3dc4b0f4859ded9b7e7a0bcf968bb9f/python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py#L116-L117

suberlak: Moved to a new task, added new tests.

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -86,6 +101,9 @@ class CalcZernikesTaskConfig(
             + "is CombineZernikesSigmaClipTask.)"
         ),
     )
```
jbkalmbach, line 103: Can you add a `doDonutStampSelector` config as well and use it to enable/disable donutStampSelector?

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -0,0 +1,197 @@
...
+from lsst.utils.timer import timeMethod
+
+
+class DonutStampSelectorTaskConfig(pexConfig.Config):
+    selectWithEntropy = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="Whether to use entropy in deciding to use the donut.",
+    )
+
```
jbkalmbach: Don't need these empty lines between each config.

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -0,0 +1,197 @@
...
+        # initiate these by selecting all donuts
+        entropySelect = np.ones(len(donutStamps), dtype="bool")
+
+        # Collect the entropy information if available
+        if "ENTROPY" in list(donutStamps.metadata):
+            entropyValue = np.asarray(donutStamps.metadata.getArray("ENTROPY"))
+            if self.config.selectWithEntropy:
+                entropySelect = entropyValue < self.config.maxEntropy
+        else:
+            self.log.warning("Entropy not in donut stamps metadata. Using all donuts.")
```
jbkalmbach: Maybe instead of "using all donuts." change to something like "No entropy cut. Checking if signal-to-noise should be applied."

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -0,0 +1,197 @@
...
+                else:
+                    snPolicyDefaults = readConfigYaml("policy:snLimitStar.yaml")
+                    filterName = donutStamps[0].bandpass
+                    filterKey = f"filter{filterName.upper()}"
+                    snThreshold = snPolicyDefaults[filterKey]
+
+                # Select using the given threshold
+                snSelect = snThreshold < snValue
+        else:
+            self.log.warning("SN not in donut stamps metadata. Using all donuts.")
```
jbkalmbach: Like above, all donuts being used isn't necessarily true if entropy cut was applied. Should change to indicate that this only means SNR cut wasn't applied.

jbkalmbach (approved):
Thanks for the changes. Looks good. Just a few comments. 

Also, one last ask. Can you run the closed loop and post a convergence plot before merging just to make sure that looks good with your default values? Maybe you've already done this.

suberlak:
Convergence plot running `img_closed_loop` with stamp selection on 
Screenshot 2024-08-28 at 10 00 57 AM

PR #225, opened 2024-03-20
Tickets/DM-43403: Instrument for full-array mode; update ComCam mask model

jfcrenshaw:
Note I must've accidentally named the branch wrong. This was meant to address [DM-43403](https://rubinobs.atlassian.net/browse/DM-43403), although we have since discovered there are more changes that need to be made to fix this issue.

[DM-43403]: https://rubinobs.atlassian.net/browse/DM-43403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/python/lsst/ts/wep/utils/taskUtils.py
+++ b/python/lsst/ts/wep/utils/taskUtils.py
@@ -264,6 +264,7 @@ def getOffsetFromExposure(
 
 def getTaskInstrument(
     camName: str,
+    detectorName: Union[str, None],
     offset: Union[float, None] = None,
     instConfigFile: Union[str, None] = None,
 ) -> Instrument:
@@ -279,6 +280,8 @@ def getTaskInstrument(
     ----------
     camName : str
         The name of the camera
+    detectorName : str or None
+        The name of the detector. Can be None.
     offset : float or None, optional
         The true offset for the exposure in mm. For LSSTCam this corresponds
         to the offset of the detector, while for AuxTel it corresponds to the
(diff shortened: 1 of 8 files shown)


python/lsst/ts/wep/utils/taskUtils.py
```diff
@@ -297,7 +300,10 @@ def getTaskInstrument(
     # Load the starting instrument
     if instConfigFile is None:
         if camName == "LSSTCam":
-            instrument = Instrument(configFile="policy:instruments/LsstCam.yaml")
+            if "_SW" in detectorName:
```
teweitsai: I would recommend to use the regular expression instead to improve the robustness. But if this is unnecessary complex, it is fine to me as well.

jmeyers314: This doesn't bother me too much since we're already under a `if camName == "LSSTCam"` predicate, but might be slightly more robust to do something like:

```python
from lsst.obs.lsst import LsstCam
from lsst.afw.cameraGeom import DetectorType
camera = LsstCam().getCamera()
if camera[detectorName].getType() == DetectorType.WAVEFRONT:
    ...
```

jmeyers314 (approved):
One minor comment; mostly just to give you an alternative if you want it.

I'm running some FAM jobs now with batoidOffsetOptic: LSSTCam hard coded (so equivalent to this branch though not _actually_ on this branch).  I'll post here when they're done as I think that's still relevant.

jmeyers314:
Hmm... My attempt to add the distortion fix resulted in the same plot as I showed you before on slack.  But since that's not on this branch anyway I'm inclined to just say merge this and let's try again on the rebased danish branch.

python/lsst/ts/wep/utils/taskUtils.py
jfcrenshaw: I used Josh's suggestion.

PR #226, opened 2024-03-30
tickets/DM-43403: fix off-axis model for full-array mode and comcam

jfcrenshaw:


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -432,9 +433,12 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         stampsMetadata["DFC_TYPE"] = np.array(
             [defocalType.value] * catalogLength, dtype=str
         )
-        stampsMetadata["DFC_DIST"] = np.array(
+        stampsMetadata["DET_OFFSET"] = np.array(
```
jbkalmbach: Keywords for FITS headers have 8 character limit.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -415,8 +415,9 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
                 detector_name=detectorName,
                 cam_name=cameraName,
                 defocal_type=defocalType.value,
-                # Save defocal offset in mm.
-                defocal_distance=instrument.defocalOffset * 1e3,
+                # Save defocal offsets in mm.
+                detector_offset=instrument.defocalOffset * 1e3,  # m -> mm
+                real_offset=instrument.batoidOffsetValue * 1e3,  # m -> mm
```
jbkalmbach: Can I suggest something like `equivalent_offset` or maybe `system_offset` instead of `real_offset`? `real_offset` isn't really clear to me what that means.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -432,9 +433,12 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         stampsMetadata["DFC_TYPE"] = np.array(
             [defocalType.value] * catalogLength, dtype=str
         )
-        stampsMetadata["DFC_DIST"] = np.array(
+        stampsMetadata["DET_OFFSET"] = np.array(
             [instrument.defocalOffset * 1e3] * catalogLength
         )
+        stampsMetadata["REAL_OFFSET"] = np.array(
```
jbkalmbach: 8 character limit as above

jbkalmbach (approved):
Looks good, just a couple small comments. Thanks. Can you hold off on merging until I run a test in the closed loop and make sure everything works and see if there's some improvement as we expect?

python/lsst/ts/wep/task/cutOutDonutsBase.py
jfcrenshaw: Yeah we can adjust this. To be clear, `real_offset` is the real offset of whatever optical element is offset (e.g. the detector, camera, or M2), while `detector_offset` is the equivalent offset of the detector. 

So I think `equivalent_offset` isn't a good name. I'm not sure `system_offset` is any more clear? What about `optic_offset`?

jbkalmbach: Ok, I had it backwards in my head I think. But yeah, `optic_offset` makes that clearer to me.

python/lsst/ts/wep/task/cutOutDonutsBase.py
jfcrenshaw: Okay I changed OFFSET -> DZ to get under character limits

python/lsst/ts/wep/task/cutOutDonutsBase.py
jfcrenshaw: Okay I changed `real_offset` -> `optic_offset`

jfcrenshaw:
Rebased on Aug 7th

PR #227, opened 2024-04-04
DM-43340: Add exposure pairing for full array mode

jmeyers314:
PR adds two new subtasks:
  `ExposurePairer` : automatically pairs intra/extra exposures based on focusz and proximity in time/space/rotation.
  `TablePairer` : pairs intra/extra exposures based on a pre-ingested table of exposure IDs.

Also included are scripts to ingest above table, and modifications to CutoutDonutsScienceSensorTask to use the pairers.

### DIFF
--- a/doc/uml/taskClass.uml
+++ b/doc/uml/taskClass.uml
@@ -29,6 +29,7 @@ CutOutDonutsScienceSensorTaskConfig *-- CutOutDonutsBaseTaskConnections
 CutOutDonutsScienceSensorTask *-- CutOutDonutsScienceSensorTaskConfig
 CutOutDonutsBaseTaskConnections <|-- CutOutDonutsScienceSensorTaskConnections
 CutOutDonutsScienceSensorTaskConfig *-- CutOutDonutsScienceSensorTaskConnections
+CutOutDonutsScienceSensorTaskConfig *-- ExposurePairer
 CutOutDonutsScienceSensorTask ..> DonutStamps
 CutOutDonutsBaseTaskConfig *-- CutOutDonutsBaseTaskConnections
 CutOutDonutsBaseTask *-- CutOutDonutsBaseTaskConfig
@@ -42,4 +43,6 @@ GenerateDonutFromRefitWcsTaskConfig *-- GenerateDonutFromRefitWcsTaskConnections
 GenerateDonutCatalogWcsTaskConfig <|-- GenerateDonutFromRefitWcsTaskConfig
 GenerateDonutFromRefitWcsTask *-- GenerateDonutFromRefitWcsTaskConfig
 GenerateDonutCatalogWcsTask <|-- GenerateDonutFromRefitWcsTask
+ExposurePairer *-- ExposurePairerConfig
+TablePairer *-- TablePairerConfig
 @enduml
--- a/python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
@@ -55,6 +55,17 @@ class CutOutDonutsScienceSensorTaskConnections(
         name="postISRCCD.visitInfo",
         multiple=True,
     )
+    donutVisitPairTable = ct.Input(
+        doc="Visit pair table",
+        dimensions=("instrument",),
+        storageClass="AstropyTable",
+        name="donutVisitPairTable",
+    )
+
+    def __init__(self, *, config=None):
+        super().__init__(config=config)
+        if not config.pairer.target._needsPairTable:
+            del self.donutVisitPairTable
 
 
 class CutOutDonutsScienceSensorTaskConfig(
@@ -110,7 +121,12 @@ class CutOutDonutsScienceSensorTask(CutOutDonutsBaseTask):
             v.dataId["visit"]: v for v in outputRefs.donutStampsExtra
         }
 
-        pairs = self.pairer.run(visitInfoDict)
+        if hasattr(inputRefs, "donutVisitPairTable"):
+            pairs = self.pairer.run(
+                visitInfoDict, butlerQC.get(inputRefs.donutVisitPairTable)
+            )
+        else:
+            pairs = self.pairer.run(visitInfoDict)
         for pair in pairs:
             exposures = butlerQC.get(
                 [exposureHandleDict[k] for k in [pair.intra, pair.extra]]
--- a/python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
@@ -29,9 +29,10 @@ import typing
 
 import lsst.afw.cameraGeom
 import lsst.afw.image as afwImage
+import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
+import lsst.pipe.base.connectionTypes as ct
 import pandas as pd
-from lsst.pipe.base import connectionTypes
 from lsst.ts.wep.task.cutOutDonutsBase import (
     CutOutDonutsBaseTask,
     CutOutDonutsBaseTaskConfig,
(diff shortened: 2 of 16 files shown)


python/lsst/ts/wep/task/pairTask.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        if self.config.doOverrideSeparation:
+            separation = self.config.overrideSeparation
+        else:
+            instrument = next(iter(visitInfos.values())).instrumentLabel
+            match instrument:
+                case "LSSTCam" | "LSSTComCam" | "LSSTComCamSim":
+                    separation = 1.5
+                case "LATISS":
+                    separation = 0.8
```
jfcrenshaw, line 113: This is neat! I didn't know about match statements

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -41,24 +42,40 @@
...
+    visitInfos = ct.Input(
         doc="Input exposure to make measurements on",
         dimensions=("exposure", "detector", "instrument"),
-        storageClass="Exposure",
-        name="postISRCCD",
+        storageClass="VisitInfo",
+        name="postISRCCD.visitInfo",
         multiple=True,
     )
+    donut_visit_pair_table = ct.Input(
```
jfcrenshaw: Everything else in this file is camelCase. It's probably better to be consistent?

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -41,24 +42,40 @@
...
 
+from .pairTask import ExposurePairer
+
 
 class CutOutDonutsScienceSensorTaskConnections(
     CutOutDonutsBaseTaskConnections, dimensions=("detector", "instrument")
 ):
-    exposures = connectionTypes.Input(
+    visitInfos = ct.Input(
         doc="Input exposure to make measurements on",
```
jbkalmbach: This docstring should be updated.

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -86,20 +107,43 @@ def runQuantum(
...
+            )
+            outputs = self.run(exposures, donutCats, camera)
+            butlerQC.put(
+                outputs.donutStampsExtra, donutStampsExtraHandleDict[pair.extra]
+            )
+            butlerQC.put(
+                outputs.donutStampsIntra,
+                donutStampsIntraHandleDict[
+                    pair.extra
+                ],  # Intentionally use extra id fo intra stamps here
```
jbkalmbach: typo "fo"

python/lsst/ts/wep/task/pairTask.py
```diff
@@ -0,0 +1,193 @@
...
+            dz = table["focusZ"] - offset
+            n_extra = np.sum(np.abs(dz) < thresh)
+            n_focal = np.sum(np.abs(dz + separation) < thresh)
+            n_intra = np.sum(np.abs(dz + 2 * separation) < thresh)
+            if n_extra + n_focal + n_intra > best_n_inliers:
+                best_n_inliers = n_extra + n_focal + n_intra
+                best_offset = offset
+
+        extraTable = table[np.abs(table["focusZ"] - best_offset) < thresh]
+        # focalTable = table[
```
jbkalmbach: Remove commented code? Or do we want to add this as an option we can use in the future?

python/lsst/ts/wep/task/pairTask.py
```diff
@@ -0,0 +1,193 @@
...
+        #     np.abs(table["focusZ"] - best_offset + separation) < thresh
+        # ]
+        intraTable = table[
+            np.abs(table["focusZ"] - best_offset + 2 * separation) < thresh
+        ]
+
+        # For each extra focal exposure, find the best intra focal exposure.
+        # Note that it's possible that the same intra exposure is paired with
+        # multiple extra exposures.
```
jbkalmbach, line 161: Do we want to allow the same intra exposure to be paired with multiple extra exposures? Would it be better to create a copy of the intra table and remove rows as they are paired?

jbkalmbach (approved):
This is really cool, thanks for adding this in! 

Can you please update the doc/uml/taskClass.uml file and the developer guide page?

Some pair tables and butler questions (these might kind of overlap but I'm just writing them down as they occurred to me reading the code):
 - How do you envision pair tables being created/stored in production? 
 - Do we want to ingest pair tables for every run into their own collection or would we have a general pair table that is regularly updated and overwritten? Are we then going to have lots of collections with just a pair table as their contents?
 - How we will know which pair tables to use when?

python/lsst/ts/wep/task/pairTask.py
jmeyers314: I left it here as potentially useful someday.  I would have left it uncommented, but then I get an unused variable flake8 error.  Happy to follow whatever advice you might have here though.

python/lsst/ts/wep/task/pairTask.py
jmeyers314: That's an interesting idea.  I'm on the fence as to whether that should be the only mode of operation, or if this should be a config setting.  Do you have a preference?

jbkalmbach: Yeah, I'm all for keeping our options open and using a config to set this.

python/lsst/ts/wep/task/pairTask.py
jbkalmbach: No problem. Maybe just a comment here mentioning that would be good.

jmeyers314:
Responses to general questions:

> How do you envision pair tables being created/stored in production?

We'll have to experiment.  It might turn out that we can mostly rely on the automatic pairer.  One option we could think about is creating a script that runs the automatic pairer and writes the result to a CSV file that we can then go and tweak by hand.  I'm happy to do that here if you like, or leave it for a future ticket.

> Do we want to ingest pair tables for every run into their own collection or would we have a general pair table that is regularly updated and overwritten? 

Yeah, that's a good question.  One pair table per collection is probably the easiest right now; I don't think butler products generally work very well with edit-in-place in mind, though you can mimic that with chain-collections where newer runs shadow older runs.

> Are we then going to have lots of collections with just a pair table as their contents?

Possibly many run-collections.  Hopefully fewer chain-collections though.

> How we will know which pair tables to use when?

I think maintaining a default pair table chain-collection may make sense here.  Quick estimate might be ~500 pairs per night and ~100 nights of commissioning.  I don't think having a 50,000 row pair table is that bad.

Off-main pairing experiments will have to maintain their own collections.  (or develop other Pairer subtasks).

jmeyers314:
Hey Bryce,
  I think I've addressed all your comments.  Do you want to take a quick peak again?  I also added a "generatePairTable.py" script to help assemble a table that can then be edited manually when using the TablePairer.  Thanks!

doc/uml/taskClass.uml
```diff
@@ -42,4 +42,6 @@ GenerateDonutFromRefitWcsTaskConfig *-- GenerateDonutFromRefitWcsTaskConnections
 GenerateDonutCatalogWcsTaskConfig <|-- GenerateDonutFromRefitWcsTaskConfig
 GenerateDonutFromRefitWcsTask *-- GenerateDonutFromRefitWcsTaskConfig
 GenerateDonutCatalogWcsTask <|-- GenerateDonutFromRefitWcsTask
+ExposurePairer *-- ExposurePairerConfig
+TablePairer *-- TablePairerConfig
```
jbkalmbach, line 47: I think `CutOutDonutsScienceSensorTaskConfig *-- ExposurePairer` needs to be added as well.

jbkalmbach (approved):
Thanks for the generate script, all my comments have been addressed. Just one comment on the UML that I think is needed based upon what we've added before.

PR #228, opened 2024-04-24
Fix blend centroid coordinates in donut stamp generation.

jbkalmbach:
Two fixes:
1. Fix donut stamp tests so that the order of operations matches that of the donutStamp code. Now the euler angle is the same in the test as in the code and not off by a negative sign.
2. Fix rotation in blend coordinates and in tests.

### DIFF
--- a/python/lsst/ts/wep/task/donutStamp.py
+++ b/python/lsst/ts/wep/task/donutStamp.py
@@ -264,13 +264,14 @@ class DonutStamp(AbstractStamp):
             # Get the offsets in the original pixel coordinates
             blendOffsets = self.blend_centroid_positions - self.centroid_position
 
-            # Rotate the coordinates to match the science sensors
-            rotMat = np.array([[0, -1], [1, 0]])
-            rotMat = np.linalg.matrix_power(rotMat, nRot)
+            # Rotate the coordinates (by -90 each time)
+            # to match the science sensors
+            rotMat = np.array([[0, 1], [-1, 0]])
+            if self.defocal_type == "extra":
+                rotMat = np.linalg.matrix_power(rotMat, nRot + 2)
+            else:
+                rotMat = np.linalg.matrix_power(rotMat, nRot)
             blendOffsets = np.transpose(rotMat @ blendOffsets.T)
-
-            # Transpose the coordinates (DVCS -> CCS)
-            blendOffsets = blendOffsets[:, ::-1]
 
         else:
             blendOffsets = None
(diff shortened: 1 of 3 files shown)


jbkalmbach:
I'm running some closed loop tests to make sure this all looks good and will post results in here for review.

python/lsst/ts/wep/task/donutStamp.py
```diff
@@ -264,14 +264,15 @@ def _setWepImage(self):
...
+            rotMat = np.array([[0, 1], [-1, 0]])
+            if self.defocal_type == "extra":
+                rotMat = np.linalg.matrix_power(rotMat, nRot + 2)
+            else:
+                rotMat = np.linalg.matrix_power(rotMat, nRot)
             blendOffsets = np.transpose(rotMat @ blendOffsets.T)
 
-            # Transpose the coordinates (DVCS -> CCS)
-            blendOffsets = blendOffsets[:, ::-1]
-
```
teweitsai, line 274: Why not use the `np.rot90()`?
https://numpy.org/doc/stable/reference/generated/numpy.rot90.html

tests/task/test_donutStamp.py
```diff
@@ -310,14 +310,18 @@ def _testWepImage(raft, sensor):
             self.assertTrue(np.allclose(wepImage.fieldAngle, center[::-1]))
 
             # Check the blend offsets are correct
-            theta = np.deg2rad(eulerZ)
+            # Rotate in opposite direction
+            theta = np.deg2rad(eulerZ) * -1.0
+            # Add extra 180 degrees for extra-focal
+            if donutStamp.defocal_type == "extra":
+                theta += np.pi
```
teweitsai, line 317: Same question above? Why not use `np.rot90()`? Maybe `-1` is np.rot90(k=2)? This might be more straight-forward to show the intention of rotation.

python/lsst/ts/wep/task/donutStamp.py
jfcrenshaw: `np.rot90` "rotates" matrices and vectors in the sense of "rolling the array of numbers onto its side". It doesn't perform a rotation in the sense of rotating a vector by a given angle.

For example, take the vector `x = np.array([[2, 1]])`. If we rotate it by 90 degrees, you get the vector `xrot = np.array([[-1, 2]])`. You can verify this is the correct rotation by taking the dot product: `np.dot(x, xrot.T)` and you will get zero.

However, if you apply `np.rot90` you get `np.array([[1], [2]])`. This is clearly not a 90 degree rotation in the usual sense, which you can verify by taking the dot product: `np.dot(x, np.rot90(x))` and you will get 4.

teweitsai: I see your point now. Thanks for the explanation!

jfcrenshaw (approved):
LGTM!

PR #229, opened 2024-04-25
Update default maxFieldDist

jbkalmbach:
Update `maxFieldDist` after analysis in DM-42067 notebook at `ts_analysis_notebooks/aos/vignetting`.


suberlak (approved):

PR #230, opened 2024-05-02
Fix units in PairerTask and add tests for ExposurePairer.

jbkalmbach:


doc/versionHistory.rst
```diff
@@ -6,14 +6,24 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-9.5.1:
+
+-------------
+9.5.1
+-------------
+
+* Fix units in PairerTask and add tests for ExposurePairer.
```
gmegh: Should it be pairTask? or should the name of the task be changed?

gmegh (approved):
LGTM! Thanks for catching that so quickly.

doc/versionHistory.rst
jbkalmbach: It's all ExposurePairer. I'll just remove reference to PairerTask.

PR #231, opened 2024-05-03
tickets/DM-44216: Fixed sharp edges on dilated, blended masks.

jfcrenshaw:


jbkalmbach (approved):
Looks good. Tested on blended objects on USDF and the edges are gone.

PR #237, opened 2024-07-09
Change CombineZernikesSigmaClipTask to use kwargs dict to set argumen…

jbkalmbach:
…ts in astropy.stats.sigma_clip.


gmegh (approved):
LGtM! Thanks

PR #238, opened 2024-07-17
Change configuration options for GenerateDonutFromRefitWcsTask to spe…

jbkalmbach:
…cify filter for photometric catalog as well.

### DIFF
--- a/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
+++ b/python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
@@ -335,11 +335,40 @@ and direct detect catalog as output."
             )
             detector = exposure.getDetector()
             filterName = exposure.filter.bandLabel
-            if self.config.photoRefFilter is not None:
-                photoRefObjLoader.config.anyFilterMapsToThis = (
-                    self.config.photoRefFilter
+            catCreateErrorMsg = (
+                "Returning new WCS but original direct detect catalog as donutCatalog."
+            )
+
+            # Check that there are catalogs
+            if len(photoRefCat) == 0:
+                self.log.warning("No catalogs cover this detector.")
+                donutCatalog = fitDonutCatalog
+                self.log.warning(catCreateErrorMsg)
+                return pipeBase.Struct(
+                    outputExposure=exposure, donutCatalog=donutCatalog
                 )
-                print(photoRefObjLoader.config)
+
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -325,6 +335,11 @@ def run(
             )
             detector = exposure.getDetector()
             filterName = exposure.filter.bandLabel
+            if self.config.photoRefFilter is not None:
+                photoRefObjLoader.config.anyFilterMapsToThis = (
+                    self.config.photoRefFilter
+                )
+                print(photoRefObjLoader.config)
```
teweitsai: Do you still need this `print()`?

python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -347,8 +362,13 @@ def run(
 
             # Except RuntimeError caused when no reference catalog
             # available for the region covered by detector
-            except RuntimeError:
-                self.log.warning("No catalogs cover this detector.")
+            except RuntimeError as refCatError:
+                if "reference filter" in str(refCatError):
```
teweitsai: Please consider to find a more reliable way to check this instead of relying on the information of `RuntimeError`. If only the error message is available, please consider to improve the way to judge to be more specific (such as regular expression with more details) instead of "reference filter" string only. Thanks!

python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -325,6 +335,39 @@ def run(
...
+            # Check that specified filter exists in catalogs
+            if self.config.photoRefFilter is not None:
+                if (
+                    f"{self.config.photoRefFilter}_flux"
+                    not in photoRefCat[0].get().schema
+                ):
+                    filterFailMsg = str(
+                        "Photometric Reference Catalog does not contain "
+                        + f"photoRefFilter: {self.config.photoRefFilter}"
+                    )
```
teweitsai: Maybe we could remove the `str` here:

```python
filterFailMsg = (
    "Photometric Reference Catalog does not contain "
    + f"photoRefFilter: {self.config.photoRefFilter}"
)
```

teweitsai (approved):
Thanks for the update!

PR #240, opened 2024-07-23
tickets/DM-45389: Removed unit config from Zernike tasks

jfcrenshaw:


tribeiro:
Are you sure you really want to do this? it is going to break a lot of existing stuff....

tribeiro (changes requested):
I would like to block this PR until we have a chance to talk about it. 

We have several systems relying on this already, this would require a lot of changing across several systems..

jfcrenshaw:
@tribeiro we had thought this would only require fixing the input units expected by `ts_ofc`, but good to know there are more problems to be aware of! @jmeyers314 @gmegh we should discuss. Depending on how problematic this is, we might just want to pin units to microns instead of nanometers.

tribeiro:
we use this on for AuxTel and some other places. If you really want to make this change (which I hope we don't) we should make sure we actually break all downstream usages by changing the name of the attribute we used to retrieve the zernikes.

Just to give you an idea of what "some other places" mean, the MTAOS publishes the estimated values. They are currently documented as microns. If we were to change this, it would require updating the MTAOS interface.

jfcrenshaw:
Assuming there's no strong preference from @jmeyers314, I think we can revert this back to microns (but we will still remove the units configuration from the task)

jmeyers314:
I would have prefered nm, but can live with microns.

tribeiro:
That would be fine by me.

jfcrenshaw:
@tribeiro can you please review the final state of this PR?

tribeiro (approved):
Looks good to me

PR #242, opened 2024-07-25
tickets/DM-43379: add single-side-of-focus mode to the TIE

jfcrenshaw:


python/lsst/ts/wep/estimation/tie.py
```diff
@@ -77,6 +78,9 @@ class TieAlgorithm(WfAlgorithm):
         between subsequent TIE iterations below which convergence is declared
         and iteration is stopped.
         (the default is 1e-9)
+    requiresPairs : bool, optional
+        Whether to allow Zernike estimation with a single donut. If False,
```
jmeyers314:
```suggestion
        Whether to allow Zernike estimation with a single donut. If True,
```

??

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -77,6 +78,9 @@ class TieAlgorithm(WfAlgorithm):
...
+        pairs are required. (the default is False)
```
jmeyers314: I'd probably make the baseline the default.  I.e., pairs are required by default.

python/lsst/ts/wep/estimation/tie.py
```diff
@@ -407,6 +429,98 @@ def history(self) -> dict:
...
+            bandLabel=refImage.bandLabel,
+            planeType="pupil",
+        )
+
+        # Create the pupil mask
+        imageMapper.createPupilMasks(pupil)
+
+        # Replace the image with the pupil mask
+        # Convolving with this Gaussian gives better results
+        pupil.image = gaussian_filter(pupil.mask.astype(float), 2)
```
jmeyers314: Does the size of the gaussian being convolved here matter?  Should we make it a config variable?

tests/estimation/test_tie.py
```diff
@@ -268,6 +269,22 @@ def testMaskBlends(self):
...
 
+    def testSingleDonut(self):
+        zkTrue, intra, extra = self._createData()
+
+        # Estimate with singles and pairs
+        tie = TieAlgorithm()
+        zkPair = tie.estimateZk(intra, extra)
+        zkIntra = tie.estimateZk(intra)
+        zkExtra = tie.estimateZk(extra)
+        diff = np.sqrt(np.sum((zkPair - np.mean([zkIntra, zkExtra], axis=0)) ** 2))
```
jmeyers314: I think we want to test `zkPair` against each of `zkIntra`, `zkExtra` instead of against their mean.

jmeyers314 (changes requested):
Structure all looks good, but I'm not sure I understand the actual algorithmic change here.  I was expecting something like:
```python
  if intra and extra:
    return TIE(intra, extra)
  elif intra:
    tmp_extra = make_image('extra', zks=0)
    return TIE(intra, tmp_extra)*2
  elif extra:
    tmp_intra = make_image('intra', zks=0)
    return TIE(tmp_intra, extra)*2
```

In particular note the factors of 2.  The intuition is that running TIE with an intra-focal donut with Zernikes zk1 and a (possibly optically inconsistent) extra-focal donut with Zernikes zk2 would yield a result of (zk1+zk2)/2.  So if we make a zk2=0 temporary donut and run with that we get (zk1 + 0)/2 out and need to multiply it by 2 to get zk1, which is what we want.  Maybe this is implicitly in the code now, but I couldn't see it.

jfcrenshaw:
> In particular note the factors of 2. The intuition is that running TIE with an intra-focal donut with Zernikes zk1 and a (possibly optically inconsistent) extra-focal donut with Zernikes zk2 would yield a result of (zk1+zk2)/2. So if we make a zk2=0 temporary donut and run with that we get (zk1 + 0)/2 out and need to multiply it by 2 to get zk1, which is what we want. Maybe this is implicitly in the code now, but I couldn't see it.

The current implementation doesn't actually simulate the other donut. Here's why:

The TIE compensates each donut individually. For each, the Zernikes used are the intrinsic Zernikes at the field angle of the donut (which is possibly unique for each donut) and the estimated aberration (which is the same for each donut). If it has the Zernikes totally correct, it should map each donut back to a uniformly illuminated pupil (which is clear because the forward mapping starts with a uniformly illuminated pupil image).

What the current implementation does is assume that the "missing" donut is just a donut that you have perfectly compensated. I.e., it says don't worry about that other donut image at all, just assume its *compensated* image is exactly a uniformly illuminated pupil. Then you can worry about estimating the Zernikes for the other image, using the perfect compensation of the other image. So if I2 is missing, it converges directly to zk1.

Does that make sense?

I did, however, think that maybe using a perfectly compensated pair effectively damps the signal of the remaining aberration. This shouldn't change the final answer, but may slow down convergence, so I tried a version where I multiplied the `zkResid` by 2. The resulting Zernikes are the same and the convergence isn't dramatically different, although it does look a little faster, so I have decided to incorporate the factor of 2.

Here is an example convergence plot for a single extrafocal donut, comparing implementations with and without the factor of 2:

image
image

It is also interesting that the max Zernike index updated by each implementation is largely different (that is what the numbers above the metrics are indicating), however it makes essentially no difference to estimated Zernikes.

python/lsst/ts/wep/estimation/tie.py
jfcrenshaw: Earlier tests I ran suggested it didn't but I double checked and I am finding that Zernike 14 is quite sensitive... not sure why, but I just made it configurable for now.

jmeyers314:
Thanks for the explanation; yes this all makes sense to me now! :)
I think you're good to go once you fix the merge conflict problems.

jmeyers314 (approved):

PR #243, opened 2024-07-31
DM-45504: Pair donuts by group dimension

jmeyers314:
Adds option to identify intra/extra pairs using the `group` butler dimension.  This requires that the group dimension was set identically for appropriate pairs when data is taken, which is already the default.

Along the way, I needed to update the dimension universe and manually set groups of some exposures of the enclosed test repo, so the gen3.sqlite3 file has some edits.  I also identified a bug in `ExposurePairer`, where AuxTel pairs were incorrectly assigned intra/extra labels.


python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
```diff
@@ -66,6 +66,11 @@ def __init__(self, *, config=None):
         super().__init__(config=config)
         if not config.pairer.target._needsPairTable:
             del self.donutVisitPairTable
+        if config.pairer.target._needsGroupDimension:
+            # Note when running with the GroupPairer you also need to change
```
jbkalmbach: Could you explain this comment a little more, I don't understand what the name of the pipeline file does when running the pipeline?

jbkalmbach:
> This requires that the group dimension was set identically for appropriate pairs when data is taken, which is already the default.

Where is this the default? In the observation scripts? Should we update `ts_imsim` to make sure we are doing this too when running future simulations because I'm not sure we're doing this there at the moment.

python/lsst/ts/wep/task/pairTask.py
```diff
@@ -248,3 +266,35 @@ def run(
...
+    _needsGroupDimension = True
+
+    def run(
+        self,
+        visitInfos: typing.Dict[int, afwImage.VisitInfo],
+    ) -> typing.List[IntraExtraIdxPair]:
+        # visitInfos should either be an intra/extra pair, or possibly an
+        # intra/extra/in-focus triplet.  Catch that here.
+        if len(visitInfos) not in (2, 3):
+            return []
```
jbkalmbach, line 303: We should add a warning or something to the log to identify the visits when this happens.

python/lsst/ts/wep/task/pairTask.py
```diff
@@ -248,3 +266,35 @@ def run(
...
+
+class GroupPairerConfig(ExposurePairerConfig):
+    def setDefaults(self):
+        super().setDefaults()
+        self.ignoreThresholds = True
+
+    def validate(self):
+        super().validate()
+
+        if not self.ignoreThresholds:
```
jbkalmbach, line 279: Can you add a test for this warning?

jbkalmbach (approved):
Looks good, just a few minor comments. Thanks!

python/lsst/ts/wep/task/cutOutDonutsScienceSensorTask.py
jmeyers314: I changed this comment to:

```
    # Note that when running with the GroupPairer you might also need to change
    # the name of the task in the pipeline yaml file (the yaml key under `tasks`
    # that points to `CutOutDonutsScienceSensorTask`) to to avoid dataset type
    # dimension conflicts with CutOutDonutsScienceSensorTask executions that
    # were run without the group pairer (and hence without the `group` input 
    # dimension).
```

jbkalmbach: Ah, okay, I get it now. Thanks.

jmeyers314:
>> This requires that the group dimension was set identically for appropriate pairs when data is taken, which is already the default.

> Where is this the default? In the observation scripts? Should we update ts_imsim to make sure we are doing this too when running future simulations because I'm not sure we're doing this there at the moment.

Right, this is in the observations scripts.  We probably should do this in `ts_imsim`.  I think it's along the lines of setting the `GROUP` field under `output.header`.  I'll look it up.

PR #244, opened 2024-08-01
tickets/DM-45524: Added utils to facilitate tests and diagnostics

jfcrenshaw:
This PR adds two utils to help with testing and making diagnostic plots.

The first is `forwardModelPair`, which forward models a pair of donuts:

```python
zkTrue, intra, extra = forwardModelPair()
```

It's highly configurable, including (but not limited to) setting Zernikes, seeing, and background noise (there are all randomly set if not provided)

The second is `plotTieConvergence` which plots the convergence of the TIE algorithm given a TIE history dictionary. For example:

image

(ignore the title on this plot, it's not present by default and I added that for something else)

The two metrics on this plot are labeled "Max" and "RMS". These are the max change in any Zernike coefficient and the RMS change of all Zernike coefficients between subsequent iterations of the TIE. There is also a number printed above the metrics for each iteration corresponding to the Noll index of the Zernike coefficient that changed the most.

### DIFF
--- a/python/lsst/ts/wep/utils/plotUtils.py
+++ b/python/lsst/ts/wep/utils/plotUtils.py
@@ -478,13 +478,20 @@ def plotMapperResiduals(
 def plotTieConvergence(history: dict, ax: Union[plt.Axes, None] = None) -> plt.Axes:
     """Plot the convergence of Zernike coefficients stored in the history.
 
-    Note that the numbers printed above the metrics for each iteration
-    are the Zernike Noll index that changed the most on that iteration.
+    The two metrics on this plot are labeled "Max" and "RMS". These are the
+    max change in any Zernike coefficient and the RMS change of all Zernike
+    coefficients between each subsequent iteration of the TIE. There is also
+    a number printed above the metrics for each iteration corresponding to the
+    Noll index of the Zernike coefficient that changed the most.
 
     Parameters
     ----------
     history : dict
-        The algorithm history dictionary
+        The algorithm history dictionary corresponding to a single Zernike
+        estimate (i.e. from a single call to the TIE solver). Note that
+        if you're loading the history from the butler metadata, you can
+        use lsst.ts.wep.utils.taskUtils.convertMetadataToHistory to convert
+        the butler format to the format expected by this function.
     ax : plt.Axes
         The matplotlib axis on which to plot the data
 
--- a/python/lsst/ts/wep/utils/plotUtils.py
+++ b/python/lsst/ts/wep/utils/plotUtils.py
@@ -25,6 +25,7 @@ __all__ = [
     "plotPupilMaskElements",
     "plotRoundTrip",
     "plotMapperResiduals",
+    "plotTieConvergence",
 ]
 
 from typing import Optional, Union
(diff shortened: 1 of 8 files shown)


python/lsst/ts/wep/utils/plotUtils.py
```diff
@@ -472,3 +473,69 @@ def plotMapperResiduals(
...
+
+def plotTieConvergence(history: dict, ax: Union[plt.Axes, None] = None) -> plt.Axes:
+    """Plot the convergence of Zernike coefficients stored in the history.
+
+    Note that the numbers printed above the metrics for each iteration
+    are the Zernike Noll index that changed the most on that iteration.
+
+    Parameters
+    ----------
+    history : dict
```
jbkalmbach, line 489: Should be clarified to point out that this takes in the history for a single pair not the full history dictionary.

jbkalmbach: Maybe also mention `convertMetadataToHistory` is available to format things correctly.

python/lsst/ts/wep/utils/plotUtils.py
```diff
@@ -472,3 +473,69 @@ def plotMapperResiduals(
     # Print info about total residuals
     print(f"mean resid: {dr[mask].mean():.3e} m")
     print(f"max resid: {dr[mask].max():.3e} m")
+
+
+def plotTieConvergence(history: dict, ax: Union[plt.Axes, None] = None) -> plt.Axes:
+    """Plot the convergence of Zernike coefficients stored in the history.
+
```
jbkalmbach, line 480: This needs a little more information. It took me a couple minutes to work through what the plot was. Mostly define what RMS and Max at each step means.

python/lsst/ts/wep/utils/testUtils.py
```diff
@@ -0,0 +1,191 @@
...
+import galsim
+import numpy as np
+from lsst.ts.wep.image import Image
+from lsst.ts.wep.imageMapper import ImageMapper
+from lsst.ts.wep.instrument import Instrument
+from scipy.ndimage import shift
+from scipy.signal import convolve
+
+
+def forwardModelPair(
```
jbkalmbach: I think this will be super useful even outside of tests! I wonder if we should rename this modelUtils or something else.

python/lsst/ts/wep/utils/testUtils.py
```diff
@@ -0,0 +1,191 @@
...
+        Zernike coefficients in meters, for Noll indices 4-jmax (inclusive)
+    Image
+        The intrafocal image
+    Image
+        The extrafocal image
+    """
+    # Create the ImageMapper that will forward model the images
+    mapper = ImageMapper(instConfig=instConfig)
+
+    # Create some random Zernikes
```
jbkalmbach: What do you think about adding an option to supply your own Zernikes? I think this method would be really nice when we need some quick sanity checks that we could do by eye on WEP output in real time.

jbkalmbach (approved):
Looks good. A few comments, mostly docstrings for the new plot utility could use a little extra information.

python/lsst/ts/wep/utils/plotUtils.py
jfcrenshaw: Done

python/lsst/ts/wep/utils/plotUtils.py
jfcrenshaw: Done.

python/lsst/ts/wep/utils/testUtils.py
jfcrenshaw: Good idea!

python/lsst/ts/wep/utils/testUtils.py
jfcrenshaw: Another great idea! This is now done

PR #245, opened 2024-08-02
tickets/DM-45524: bug fix and random field angles

jfcrenshaw:
Fixed bug and added random field angles in lsst.ts.wep.utils.modelUtils.forwardModelPair

### DIFF
--- a/python/lsst/ts/wep/utils/modelUtils.py
+++ b/python/lsst/ts/wep/utils/modelUtils.py
@@ -43,8 +43,8 @@ def forwardModelPair(
     seeing: Union[float, None] = None,
     skyLevel: Union[float, None] = None,
     bandLabel: str = "r",
-    fieldAngleIntra: tuple = (0.23, -0.6),
-    fieldAngleExtra: tuple = (0.23, -0.6),
+    fieldAngleIntra: Union[tuple, None] = None,
+    fieldAngleExtra: Union[tuple, None] = None,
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/utils/modelUtils.py
```diff
@@ -88,11 +88,15 @@ def forwardModelPair(
...
+        only specified for the intra or extrafocal donut, both donuts use
+        that angle. If neither is specified, a random angle is used for
+        both. (the default is None)
     fieldAngleExtra : tuple, optional
-        Field angle, in degrees, of the extrafocal donut.
-        (the default is (0.23, -0.6))
+        Field angle, in degrees, of the extrafocal donut. If the angle is
+        only specified for the intra or extrafocal donut, both donuts use
+        that angle. If neither is specified, a random angle is used for
+        both. (the default is None)
```
jbkalmbach: Should specify that it's the same random angle for both in the last case. It's not quite clear.

jbkalmbach (changes requested):
This new function needs some tests so we don't miss any more things like this.

python/lsst/ts/wep/utils/modelUtils.py
jfcrenshaw: Done.

jbkalmbach (approved):
Looks good. Thanks for adding the tests.

PR #246, opened 2024-08-02
Separate recenterFlags in cutOutDonuts tasks metadata into recenterFlagsExtra and recenterFlagsIntra.

jbkalmbach:


jfcrenshaw (approved):
LGTM in principle, but there are failed tests that need to be addressed

PR #248, opened 2024-08-06
Tickets/DM-45279: add visit to donutStamps metadata

suberlak:
To help with plotAOSTask, this allows to find out what was the exposure (visit) used to cut out donuts from, regardless of which visit they are stored with (as currently both `donutStampsExtra` and `donutStampsIntra` get saved with extra-focal `visit`, this allows to retrieve that at a later stage if we want to pass `donutStampsIntra` with the intra-focal visit).

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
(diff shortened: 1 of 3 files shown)


jmeyers314 (approved):
LGTM!  Thanks!

PR #250, opened 2024-08-12
Add pipelines directory to easily share pipeline templates.

jbkalmbach:


gmegh (approved):
LGTM. Just one question, ComCam does not need anything different to run than LSSTCam, right? I am pretty sure that's right, but I want to double-check just in case we are missing something. Thanks for being so quick!

jbkalmbach:
Just need to change the instrument but the pipeline is the same.

PR #252, opened 2024-08-14
Add comcam full default pipeline

gmegh:


pipelines/fullPipelines/comcamScienceSensorFullPipeline
```diff
@@ -0,0 +1,75 @@
+description: wep basic processing test pipeline
+instrument: lsst.obs.lsst.LsstComCam
+tasks:
+  generateDonutDirectDetectTask:
+    class: lsst.ts.wep.task.generateDonutDirectDetectTask.GenerateDonutDirectDetectTask
+    config:
+      donutSelector.useCustomMagLimit: True
+      donutSelector.sourceLimit: 5
+  cutOutDonutsScienceSensorTask:
```
jmeyers314: Let's use the group pairer. See https://github.com/lsst-ts/ts_wep/blob/develop/tests/testData/pipelineConfigs/testCutoutsFamPipelineGroupPairer.yaml#L6 for an example.

pipelines/fullPipelines/comcamScienceSensorFullPipeline
```diff
@@ -0,0 +1,75 @@
...
+      doDefect: False
+      doNanMasking: False
+      doInterpolate: False
+      doBrighterFatter: False
+      doDark: False
+      doFlat: False
+      doApplyGains: True
+      doFringe: False
+      doOverscan: True
+      python: OverscanCorrectionTask.ConfigClass.fitType = 'MEDIAN'
```
jmeyers314: @mfisherlevine , how do these ISR settings compare to the quicklook pipeline?

pipelines/fullPipelines/comcamScienceSensorFullPipeline
```diff
@@ -0,0 +1,75 @@
...
+      - aggregateZernikesTask
+      - aggregateDonutCatalogsTask
+      - aggregateAOSVisitTableTask
+      - plotAOSTask
+      - aggregateDonutStampsTask
+      - plotDonutTask
+    description: |
+      AOS Donut visualization plotting tasks. This step generates plots
+      (including the pyramid residual and donut gallery) and
+      tables for the AOS visit.
```
jbkalmbach: Add blank line at end please.

jmeyers314 (changes requested):
Let's change this to something like comCamRapidAnalysisPipeline.yaml

pipelines/fullPipelines/comcamScienceSensorFullPipeline
```diff
@@ -0,0 +1,75 @@
+description: wep basic processing test pipeline
```
jbkalmbach: Can we change the directory name to something more descriptive like "production"?

pipelines/fullPipelines/comcamScienceSensorFullPipeline
mfisherlevine: Errr, very different indeed, to be honest! quicklook is basically the same as DRP expect with `doBrighterFatter` set to `False` because it's slow, but everything else is super quick and also necessary for pretty much all processing. So the good news is that you'd be getting all of that for free, the bad news is that it's nothing like your current setup...

jmeyers314: Okay, I vote we adopt the quicklook settings.  Guillem, I think you can copy those from https://github.com/lsst/drp_pipe/blob/main/pipelines/_ingredients/DRP-full.yaml#L12 and https://github.com/lsst/drp_pipe/blob/main/pipelines/LSSTComCam/quickLook.yaml#L8.

mfisherlevine: I _think_ you should be able to import them actually, rather than copying them. But then maybe this would need to live in that package, not sure. I'd have no problem with that being the case though, personally.

pipelines/fullPipelines/comcamScienceSensorFullPipeline
gmegh: Great suggestion, i really didn't like fullPipelines but ran out of naming imagination

pipelines/fullPipelines/comcamScienceSensorFullPipeline
gmegh: Out of curiosity, what does this do, exactly?

jmeyers314: It controls how exposures are paired up when you run in batch mode (this "pairer" looks at the value of the "group" dimension to find pairs).  It's also required right now if you want to append new aos pipeline results to an existing output collection.

jmeyers314:
LGTM.  Thanks Guillem!

jmeyers314 (approved):

pipelines/production/comCamRapidAnalysisPipeline.yaml
```diff
@@ -0,0 +1,73 @@
+description: rapid analysis pipeline for ComCam
+instrument: lsst.obs.lsst.LsstComCam
+tasks:
+  generateDonutDirectDetectTask:
```
jbkalmbach, line 4: Sorry, I know this has been hanging around a while but while preparing our notebooks for analysis I've been thinking about`generateDonutFromRefitWcsTask` as our default rather than `generateDonutDirectDetectTask`. What do you think? `fromRefitWcs` might not work if the pointing model is really bad at the start but I think we should talk about trying to use it. Anyway, maybe something to bring up during an AOS meeting.

PR #253, opened 2024-08-16
Tickets/DM-45806: Fix generateDonutDirectDetect with no sources

suberlak:
Handle correctly case of no sources detected, improve the tests.

### DIFF
--- a/tests/task/test_generateDonutDirectDetectTask.py
+++ b/tests/task/test_generateDonutDirectDetectTask.py
@@ -20,8 +20,10 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import os
-import unittest
+from copy import copy
 
+import lsst.afw.image as afwImage
+import lsst.utils.tests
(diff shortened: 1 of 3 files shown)


tests/task/test_generateDonutDirectDetectTask.py
```diff
@@ -103,46 +149,106 @@ def testUpdateDonutCatalog(self):
...
+            "coord_dec",
+            "centroid_x",
+            "centroid_y",
+            "detector",
+            "source_flux",
+            "blend_centroid_x",
+            "blend_centroid_y",
+        ]
+        self.assertCountEqual(taskOutNoSrc.donutCatalog.columns, expected_columns)
+
```
jbkalmbach, line 199: I think this should be broken up into two separate tests. One like you say in your comments `testTaskRunNormal` and another `testTaskRunNoSrc`.

jbkalmbach (approved):
Looks good, thanks!

PR #255, opened 2024-08-28
Subtract median shifts from stamps before applying recentering cut.

jbkalmbach:


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -167,12 +167,12 @@ def __init__(self, **kwargs):
...
         self.maxRecenterDistance = self.config.maxRecenterDistance
 
-    def shiftCenter(self, center, boundary, distance):
+    def shiftCenters(self, centerArr, boundary, distance):
         """Shift the center if its distance to boundary is less than required.
 
         Parameters
         ----------
-        center : float
+        centerArr : float
```
suberlak: So when I first look  the name seems to imply an array, yet it seems to be the float. Which type is it?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -167,12 +167,12 @@ def __init__(self, **kwargs):
         # Set max recentering distance in pixels
         self.maxRecenterDistance = self.config.maxRecenterDistance
 
-    def shiftCenter(self, center, boundary, distance):
+    def shiftCenters(self, centerArr, boundary, distance):
```
suberlak, line 187: Name implies multiple centers,  so if I understand it correctly the inline explanation should point to the same (?)

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -186,13 +186,14 @@ def shiftCenter(self, center, boundary, distance):
         """
 
         # Distance between the center and boundary
```
suberlak: See above - is it a single center or multiple?

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -345,53 +382,55 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         yCornerList = list()
 
         # Keep track of recentering failures
-        recenterFlags = list()
+        recenterFlags = np.zeros(len(donutCatalog), dtype=int)
+        xCent = donutCatalog["centroid_x"].values
```
suberlak: It would be easier to follow with a semantics that reflects that here we have an array of all donut centers, so eg.  `xCenters`  ...

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -345,53 +382,55 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
+        yCent = donutCatalog["centroid_y"].values
+        finalDonutX, finalDonutY, xCorner, yCorner, initXCorner, initYCorner = (
+            self.calculateFinalCentroid(exposure, template, xCent, yCent)
+        )
+        xShift = finalDonutX - xCent
```
suberlak: Since this is also a vector now it would be clearer with eg.`xShifts`

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -345,53 +382,55 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
 
             # Get the final cutout
-            finalCorner = lsst.geom.Point2I(xCorner, yCorner)
+            finalCorner = lsst.geom.Point2I(donutRow["xCorner"], donutRow["yCorner"])
             finalBBox = lsst.geom.Box2I(
                 finalCorner, lsst.geom.Extent2I(self.donutStampSize)
             )
-            xCornerList.append(xCorner)
-            yCornerList.append(yCorner)
+            xCornerList.append(donutRow["xCorner"])
```
suberlak: So now this can be done outside the loop, since all `xCorners` are known... But in fact since the only reason for this list was to store in the loop the values that are later stored as array in `stampsMetadata`, I think that  entire `xCornerList` can be removed.

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -345,53 +382,55 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         yCornerList = list()
```
suberlak: As mentioned below, I think that both `xCornerList` and `yCornerList` can be safely removed just like `finalXCentList`..

tests/task/test_cutOutDonutsBase.py
```diff
@@ -173,17 +174,19 @@ def testCalculateFinalCentroid(self):
             offCenterExp,
             offCenterCoord,
         ) = self._generateTestExposures()
+        centerCoordX = np.array([centerCoord[0]])
```
suberlak: So I think this only tests it for an array of a single value eg [1].  Would it make sense to also test that the new vectorized behavior meant to recenter multiple donuts at the same time works as expected?

suberlak (approved):
Thanks for the update!

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: Good catch.

tests/task/test_cutOutDonutsBase.py
jbkalmbach: Yep, added multiple sources to test arrays.

PR #258, opened 2024-09-04
Change CalcZernikesTask output to be at least 2D for average as well …

jbkalmbach:
…as raw to make integration with MTAOS easier.


gmegh (approved):
LGTM! Thanks!

PR #260, opened 2024-09-18
DM-46393: Optionally bin donut stamps before estimating wavefront

jmeyers314:
Adds config option to both TIE and danish to bin the donut stamps before estimating the wavefront.  In some spot checks, the accuracy difference appears negligible while the speed up is ~3x for TIE and ~2x for danish.

### DIFF
--- a/python/lsst/ts/wep/instrument.py
+++ b/python/lsst/ts/wep/instrument.py
@@ -19,6 +19,8 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+from __future__ import annotations
+
 __all__ = ["Instrument"]
 
 from functools import lru_cache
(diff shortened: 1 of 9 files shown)


python/lsst/ts/wep/instrument.py
```diff
@@ -162,6 +164,23 @@ def __init__(
         # Check the config
         self.checkConfig()
 
+    def clone(self) -> Instrument:
```
jfcrenshaw: The `Image` object has a very similar method called `Image.copy()`. Maybe also name this copy so the names match?

jfcrenshaw (approved):
Looks great! Two small comments, and it looks like you need to update the version history.

Once you merge this, feel free to immediately open a PR to main, and I will approve that too.

python/lsst/ts/wep/instrument.py
```diff
@@ -162,6 +164,23 @@ def __init__(
...
+        """Return a deep copy of the instrument."""
```
jfcrenshaw: The one thing this misses is that the new instrument object will be missing the caches of the original instrument object. This shouldn't be a big overhead hit, but maybe you should mention the caches in the docstring here, in case in the future a `copy()` call ends up inside a loop or something

PR #261, opened 2024-09-18
Adjust masking used in calculateSN to only use the donut masked pixel…

jbkalmbach:
…s and add limit on dilation to avoid infinite loop when checking if mask goes to edge.

### DIFF
--- a/python/lsst/ts/wep/task/cutOutDonutsBase.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsBase.py
@@ -183,6 +183,8 @@ class CutOutDonutsBaseTask(pipeBase.PipelineTask):
         self.makeSubtask("subtractBackground")
         # Set max recentering distance in pixels
         self.maxRecenterDistance = self.config.maxRecenterDistance
+        # Set Variance Plane Warning only once
+        self.varianceWarningSet = False
 
     def shiftCenters(self, centerArr, boundary, distance):
         """Shift the centers of sources if the distance to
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -352,78 +353,81 @@ def calculateSN(self, stamp):
         # Thus to find out the number of pixels
         # taken by the donut mask we sum all
         # the nonzero mask pixels.
-        donut_mask = stamp.stamp_im.mask.array > 0
+        donutMaskPlane = mask.getMaskPlane("DONUT")
+        donutMask = mask.array == np.power(2, donutMaskPlane)
```
suberlak, line 359: Is this tested behavior? Making sure that this approach yields the same number of donut pixels as before

suberlak (approved):
Thanks for the update!

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: Yes, I've tested as shown here. Without blends the behavior is the same.

But I guess one question is what to do about blended areas? Do you want to include them or avoid them when calculating signal? The calculated signal will be higher in the blended areas than the true value for the central source leading to a higher overall SNR than the true value. Should we exclude these areas then or keep them?

Single Donut:
Number of pixels in each test case: 9481
image

Donut w/ blend the way it is now cuts out the blended area when calculating donut mask:
image

suberlak: Using the updated names, for `ttlSignalMean` and `ttlSignalSum` we use the default donut mask since the mask usually goes up to the donut edge (rarely much beyond the edge), and the point of dilating the donut mask (and then inverting it to select the background region) was to provide a backup method of estimating background variance: `ttlNoiseBkgndVariance` is only used if `ttlNoiseDonutVariance` is not available (in case of no variance plane). But in that edge case (working with old dataset containing blends), it would be better to use the donut mask with all pixels for the background variance, because only then it contains the actual background. For instance, here is an example of very dense region 
image

suberlak: If the brightest background donut (towards the top of the postage stamp) were included in the mask as the "blended mask",  then the variance of background would decrease. As it is, the region outside of the dilated donut mask has high variance dominated by the signal of these background donuts (which is why we would ordinarily reject it as this would decrease the S/N). So for a recognized blend, using the combined "blend" mask would decrease the estimated background variance, which would increase the S/N.

PR #265, opened 2024-09-25
Fixed bug in donutSourceSelectorTask where the task set with maxBlend…

jbkalmbach:
…ed > 0 and sources with a number of overlapping donuts greater than maxBlended did not give correct blend centers in the final catalog.

### DIFF
--- a/python/lsst/ts/wep/task/donutSourceSelectorTask.py
+++ b/python/lsst/ts/wep/task/donutSourceSelectorTask.py
@@ -321,12 +321,12 @@ class DonutSourceSelectorTask(pipeBase.Task):
                     # all the faint sources that we don't care about
                     # when deblending. Add one to index because
                     # magDiff is all sources in magSortedDf after index=0.
-                    maxIdx = np.where(magDiff < minMagDiff)[0] + 1
+                    blendMagIdx = np.where(magDiff < minMagDiff)[0] + 1
                     blendCentersX[groupIndices[srcOn]] = (
-                        magSortedDf["x"].iloc[idxList[maxIdx]].values
+                        magSortedDf["x"].iloc[idxList[blendMagIdx]].values
                     )
                     blendCentersY[groupIndices[srcOn]] = (
-                        magSortedDf["y"].iloc[idxList[maxIdx]].values
+                        magSortedDf["y"].iloc[idxList[blendMagIdx]].values
                     )
                     sourcesKept += 1
                 # Keep the source if it is blended with up to maxBlended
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/task/donutSourceSelectorTask.py
```diff
@@ -336,11 +337,14 @@ def selectSources(self, sourceCat, detector, filterName):
...
+                    # Same process as above to make sure we only get
+                    # the blend centers we care about
+                    maxIdx = np.where(magDiff < minMagDiff)[0] + 1
                     blendCentersX[groupIndices[srcOn]] = (
-                        magSortedDf["x"].iloc[idxList[1 : maxBlended + 1]].values
+                        magSortedDf["x"].iloc[idxList[maxIdx]].values
                     )
                     blendCentersY[groupIndices[srcOn]] = (
-                        magSortedDf["y"].iloc[idxList[1 : maxBlended + 1]].values
+                        magSortedDf["y"].iloc[idxList[maxIdx]].values
```
suberlak: Previously the blends provided a list, and now it is a single `maxIdx` ?

tests/task/test_donutSourceSelectorTask.py
```diff
@@ -60,8 +60,8 @@ def _createTestCat(self):
         self.magMin = magPolicyDefaults[defaultFilterKey]["low"]
         self.magRange = self.magMax - self.magMin
         magArray = [
-            self.magMax + 1.0,
             self.magMax - 0.1,
+            self.magMax + 1.0,
```
suberlak, line 64: This looks like just reordering from `black`, not a structural change?

tests/task/test_donutSourceSelectorTask.py
```diff
@@ -320,6 +320,18 @@ def testSelectSources(self):
         testCatSelected = testCatStruct.selected
         self.assertListEqual(list(testCatSelected), [False, False, True, True])
 
+        # Test that when there are multiple blends but no all are within
```
suberlak: typo: "not all"

tests/task/test_donutSourceSelectorTask.py
```diff
@@ -320,6 +320,18 @@ def testSelectSources(self):
...
+        # isolatedMagDiff that the correct donuts are
```
suberlak: I think unfinished comment sentence?

suberlak (approved):
Thanks for the update!

tests/task/test_donutSourceSelectorTask.py
jbkalmbach: No, this was an intentional change. When I reorder the donuts like this then the closest donut gets added to the catalog when `maxBlended` = 1 with the old code even though it is the fainter donut and outside the `isolatedMagDiff` limit and shouldn't count as blended. This makes the old code fail the new test I've added at the end. So reordering allows us to check this behavior and the new code which fixes it then passes the new test.

python/lsst/ts/wep/task/donutSourceSelectorTask.py
jbkalmbach: Still a list, renamed this to `blendMagIdx` to clarify that these are the sources with magnitudes bright enough to count as blended.

tests/task/test_donutSourceSelectorTask.py
jbkalmbach: Cleaned up, thanks.

PR #267, opened 2024-10-04
Change from pandas dataframe in generateDonutCatalogUtils to astropy …

jbkalmbach:
…table.

### DIFF
--- a/python/lsst/ts/wep/task/cutOutDonutsBase.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsBase.py
@@ -66,8 +66,8 @@ class CutOutDonutsBaseTaskConnections(
             "detector",
             "instrument",
         ),
-        storageClass="DataFrame",
-        name="donutCatalog",
+        storageClass="AstropyQTable",
+        name="donutCatalogAstropy",
         multiple=True,
     )
     camera = connectionTypes.PrerequisiteInput(
@@ -477,7 +477,7 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
         ----------
         exposure : lsst.afw.image.Exposure
             Post-ISR image with defocal donuts sources.
-        donutCatalog : pandas.DataFrame
+        donutCatalog : astropy.table.QTable
             Source catalog for the pointing.
         defocalType : enum 'DefocalType'
             Defocal type of the donut image.
@@ -495,6 +495,7 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
         detectorName = detector.getName()
         bandLabel = exposure.filter.bandLabel
         visitId = exposure.getInfo().getVisitInfo().id
+        catalogMeta = donutCatalog.meta
 
         # Run background subtraction
         self.subtractBackground.run(exposure=exposure).background
@@ -533,8 +534,8 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
 
         # Keep track of recentering failures
         recenterFlags = np.zeros(len(donutCatalog), dtype=int)
-        xCenters = donutCatalog["centroid_x"].values
-        yCenters = donutCatalog["centroid_y"].values
+        xCenters = donutCatalog["centroid_x"].value
+        yCenters = donutCatalog["centroid_y"].value
         (
             finalDonutX,
             finalDonutY,
@@ -571,7 +572,7 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
         # Value of entropy
         stampsEntropy = list()
 
-        for donutRow in donutCatalog.to_records():
+        for idx, donutRow in enumerate(donutCatalog):
             # Make an initial cutout larger than the actual final stamp
             # so that we can centroid to get the stamp centered exactly
             # on the donut
@@ -585,7 +586,7 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
                     "Donut Recentering Failed. Flagging and "
                     + f"not shifting center of stamp for {defocalType.value}-focal "
                     + f"source at {(donutRow['centroid_x'], donutRow['centroid_y'])}. "
-                    + f"Catalog index: {donutRow.index}. "
+                    + f"Catalog row: {donutRow.index}. "
                     + f"Proposed Shift: {(donutRow['xShift'], donutRow['yShift'])}."
                 )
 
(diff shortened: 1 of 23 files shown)


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -689,13 +691,13 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
         # Save the donut flux as magnitude
         if len(donutCatalog["source_flux"]) > 0:
```
suberlak, line 692: does this line need a `.value` too?

python/lsst/ts/wep/task/generateDonutCatalogUtils.py
```diff
@@ -73,7 +76,7 @@ def runSelection(refObjLoader, detector, wcs, filterName, donutSelectorTask):
         )
 
 
-def donutCatalogToDataFrame(
+def donutCatalogToAstropy(
     donutCatalog=None, filterName=None, blendCentersX=None, blendCentersY=None
 ):
     """
```
suberlak, line 82: Would need to switch the docstring for this method too from pandas to astropy

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
```diff
@@ -75,8 +76,8 @@ class GenerateDonutCatalogWcsTaskConnections(
             "detector",
             "instrument",
         ),
-        storageClass="DataFrame",
-        name="donutCatalog",
+        storageClass="AstropyQTable",
+        name="donutCatalogAstropy",
```
suberlak: given that we're not keeping `donutCatalogDataFrame` alongside this one,  would it not be easier to retain `donutCatalog` name? This would then not break the compatibility with existing code ...

python/lsst/ts/wep/task/generateDonutFromRefitWcsTask.py
```diff
@@ -104,8 +104,8 @@ class GenerateDonutFromRefitWcsTaskConnections(
             "detector",
             "instrument",
         ),
-        storageClass="DataFrame",
-        name="donutCatalog",
+        storageClass="AstropyQTable",
+        name="donutCatalogAstropy",
```
suberlak: as above, what do you think about keeping `donutCatalog` ?

tests/task/test_cutOutDonutsBase.py
```diff
@@ -323,7 +323,7 @@ def _getExpAndCatalog(self, defocalType):
             "postISRCCD", dataId=dataId, collections=[self.runName]
         )
         donutCatalog = self.butler.get(
-            "donutCatalog", dataId=dataId, collections=[self.runName]
+            "donutCatalogAstropy", dataId=dataId, collections=[self.runName]
```
suberlak: for this reason (keeping `butler.get('donutCatalog'`) I think we should not change that.

tests/task/test_cutOutDonutsCwfsTask.py
```diff
@@ -132,10 +132,10 @@ def _getDataFromButler(self):
         )
         # Get the donut catalogs for each detector
         donutCatalogExtra = self.butler.get(
-            "donutCatalog", dataId=self.dataIdExtra, collections=[self.runName]
+            "donutCatalogAstropy", dataId=self.dataIdExtra, collections=[self.runName]
```
suberlak: as above, I'd recommend keeping `donutCatalog` since we're not going to have dual `donutCatalogDataFrame` and `donutCatalogAstropy` side by side - we're only switching to a new standard.

tests/task/test_cutOutDonutsCwfsTask.py
```diff
@@ -132,10 +132,10 @@ def _getDataFromButler(self):
...
         )
         donutCatalogIntra = self.butler.get(
-            "donutCatalog", dataId=self.dataIdIntra, collections=[self.runName]
+            "donutCatalogAstropy", dataId=self.dataIdIntra, collections=[self.runName]
```
suberlak: as above, I'd keep unchanged for compatibility

tests/task/test_cutOutDonutsScienceSensorTask.py
```diff
@@ -233,10 +233,10 @@ def testTaskRun(self):
         )
 
         donutCatalogExtra = self.butler.get(
-            "donutCatalog", dataId=self.dataIdExtra, collections=[self.runName]
+            "donutCatalogAstropy", dataId=self.dataIdExtra, collections=[self.runName]
```
suberlak: As above

tests/task/test_cutOutDonutsScienceSensorTask.py
```diff
@@ -233,10 +233,10 @@ def testTaskRun(self):
...
         )
         donutCatalogIntra = self.butler.get(
-            "donutCatalog", dataId=self.dataIdIntra, collections=[self.runName]
+            "donutCatalogAstropy", dataId=self.dataIdIntra, collections=[self.runName]
```
suberlak: As above

tests/task/test_generateDonutCatalogWcsTask.py
```diff
@@ -153,41 +153,43 @@ def testPipeline(self):
             },
             collections=[f"{runName}"],
         )
-        donutCatDf_S11 = pipelineButler.get(
-            "donutCatalog",
+        donutCatTable_S11 = pipelineButler.get(
+            "donutCatalogAstropy",
```
suberlak: as above, it would make it less disruptive if we kept `donutCatalog`

tests/task/test_generateDonutCatalogWcsTask.py
```diff
@@ -153,41 +153,43 @@ def testPipeline(self):
...
             dataId={"instrument": "LSSTCam", "detector": 94, "visit": exposureId},
             collections=[f"{runName}"],
         )
-        donutCatDf_S10 = pipelineButler.get(
-            "donutCatalog",
+        donutCatTable_S10 = pipelineButler.get(
+            "donutCatalogAstropy",
```
suberlak: as above, it would make it less disruptive if we kept `donutCatalog`

suberlak (approved):
Thanks for the update! Just small fixes to docstrings, and see comments about `donutCatalog`  butler data type.

python/lsst/ts/wep/task/generateDonutCatalogWcsTask.py
jbkalmbach: We have to use a different name since `donutCatalog` is already defined in the butler as a `DataFrame`.

python/lsst/ts/wep/task/cutOutDonutsBase.py
jbkalmbach: no, len() will work just fine here with or without. I'm just finding the length of the catalog.

PR #273, opened 2024-10-18
tickets/DM-46117: single-side-of-focus pipeline

jfcrenshaw:


python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -0,0 +1,251 @@
...
+        storageClass="NumpyArray",
+        name="zernikeEstimateRaw",
+    )
+    outputZernikesAvg = connectionTypes.Output(
+        doc="Zernike Coefficients averaged over donuts",
+        dimensions=("visit", "detector", "instrument"),
+        storageClass="NumpyArray",
+        name="zernikeEstimateAvg",
+    )
+    zernikes = connectionTypes.Output(
```
jbkalmbach, line 60: Should be a QTable now.

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
```diff
@@ -0,0 +1,251 @@
...
+            zkTable.add_row(row)
+
+        zkTable.meta["intra"] = {}
+        zkTable.meta["extra"] = {}
+
+        for dict_, stamps in [
+            (zkTable.meta["intra"], intraStamps),
+            (zkTable.meta["extra"], extraStamps),
+        ]:
+            if len(stamps) > 0:
```
jbkalmbach: It would actually make more sense to break this off into a separate function that is used here and in `calcZernikesTask` since we're using it in two places now and we don't want to have to update things twice every time.

python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -74,7 +74,8 @@ class EstimateZernikesBaseConfig(pexConfig.Config):
         default=True,
         doc="Whether to estimate Zernike coefficients from pairs of donut "
         + "stamps. If False, Zernikes are estimated from individual donuts. "
-        + "Note the TIE algorithm requires pairs of donuts.",
+        + "Note if ONLY intra- or extra-focal stamps are passed, then Zernikes "
```
jbkalmbach: This is confusing. If `usePairs` is `True` and there are not pairs of donuts it should warn you and refuse to run instead of running silently like this.

jbkalmbach (approved):
Looks good. Just a couple comments. Biggest one would be creating new function to help improve maintainability.

Can you update the doc/developer-guide as well?

python/lsst/ts/wep/task/estimateZernikesBase.py
jfcrenshaw: I removed this ambiguity by removing `usePairs`. It wasn't really necessary anymore, since unpaired estimation is now its own task/pipeline.

python/lsst/ts/wep/task/calcZernikesUnpairedTask.py
jfcrenshaw: Good idea. I put this in a method in `calcZernikesTask`, but made it general enough that the unpaired task could use it too.

PR #277, opened 2024-10-26
tickets/DM-47163: Cut on fraction of bad pixels; run stamp selection during CalcZernikeTask by default

jfcrenshaw:


python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -672,6 +675,10 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
             isEffective.append(eff)
             stampsEntropy.append(entro)
 
+            # Calculate fraction of bad pixels
+            bits = finalStamp.mask.getPlaneBitMask(("SAT", "BAD", "NO_DATA"))
```
suberlak: We could set this dictionary of default mask plane bits as a config here in `cutOutDonuts`, eg.  `badPixelMaskDefinitions`, so that if someone wants to consider some other values in the future it's easy to do

suberlak: Eg https://github.com/lsst/ip_isr/blob/main/python/lsst/ip/isr/isrFunctions.py#L182  and `CR` may be a choice one may use too  , so we  ould have that the default is `"SAT", "BAD", "NO_DATA"`

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -62,13 +68,17 @@ class DonutStampSelectorTaskConfig(pexConfig.Config):
         default=3.5,
         doc=str("The entropy threshold to use (keep donuts only below the threshold)."),
     )
+    maxFracBadPixels = pexConfig.Field(
+        dtype=float,
+        default=0.0,
```
suberlak, line 73: It would be really good to run that against even a simulation   / early real data. I'm worried that with such strict requirement we may get nothing passing atm.

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -194,12 +201,27 @@ def selectStamps(self, donutStamps):
...
+
+        # collect fraction-of-bad-pixels information if available
+        if "FRAC_BAD_PIX" in list(donutStamps.metadata):
+            fracBadPix = np.asarray(donutStamps.metadata.getArray("FRAC_BAD_PIX"))
+            if self.config.selectWithFracBadPixels:
+                fracBadPixSelect = fracBadPix <= self.config.maxFracBadPixels
+        else:
+            self.log.warning("No fraction-of-bad-pixels cut.")
+
         # AND condition : if both selectWithEntropy
```
suberlak: Since we're adding another condition we can probably also fix that comment (eg. `AND statement: choose only donuts that satisfy all selected conditions`)

suberlak (approved):
Thanks for the update!

python/lsst/ts/wep/task/donutStampSelectorTask.py
jfcrenshaw: Nice catch!

python/lsst/ts/wep/task/donutStampSelectorTask.py
jfcrenshaw: Working on this

python/lsst/ts/wep/task/cutOutDonutsBase.py
jfcrenshaw: Good idea!

jfcrenshaw:
New failures arose after [Adding frac_bad_pix to calcZernikes output.](https://github.com/lsst-ts/ts_wep/pull/277/commits/b36ca26c07503c5d1616c5836780404d644f52aa)

Also tests on ComCam data revealed that some pixel fractions are > 1, which shouldn't happen. Will investigate soon.

python/lsst/ts/wep/task/donutStampSelectorTask.py
jfcrenshaw: Checked this works

PR #278, opened 2024-10-27
tickets/DM-47164: Update to new ISR task

jfcrenshaw:


jbkalmbach:
I think we can close this PR? We have updated the ISR task in the pipeline files and cleaned them up a bit as well so I think this is all out of date now.

PR #279, opened 2024-10-28
Tickets/DM-47188: Fix bug where CalcZernikesTask fails if number of intra/extra stamps isn't equal

jfcrenshaw:


tests/task/test_calcZernikesTieTaskCwfs.py
```diff
@@ -269,3 +269,15 @@ def testWithAndWithoutPairs(self):
...
+        # Test for when you have more of either extra or intra
+        # Load the test data
+        stampsExtra = self.donutStampsExtra
+        stampsIntra = self.donutStampsIntra
+
+        # Increase length of extra list
+        stampsExtra.extend([stampsExtra[0]])
+
+        # Now estimate Zernikes
+        self.task.run(stampsExtra, stampsIntra)
```
suberlak, line 283: So if it's supposed to stop can you test that indeed it's not going further?  Or what would have happened if `break` wasn't there to test the new functionality.

python/lsst/ts/wep/task/donutStampSelectorTask.py
```diff
@@ -172,8 +172,10 @@ def selectStamps(self, donutStamps):
         entropySelect = np.ones(len(donutStamps), dtype="bool")
 
         # Collect the entropy information if available
+        entropyValue = np.full(len(donutStamps), np.nan)
         if "ENTROPY" in list(donutStamps.metadata):
-            entropyValue = np.asarray(donutStamps.metadata.getArray("ENTROPY"))
+            fillVals = np.asarray(donutStamps.metadata.getArray("ENTROPY"))
```
suberlak, line 177: is this helping with the case where most of the donutStamps do not have entropy, but only some do?  I'm wondering about the ordering (`entropyValue` has the same length as `donutCatalog` that corresponds to cutout `donutStamps`,  so why would `fillVals` be shorter than that?

python/lsst/ts/wep/task/calcZernikesTask.py
```diff
@@ -238,6 +238,12 @@ def createZkTable(
                 zkCoeffCombined.flags,
             )
         ):
+            # If zk is None, we need to stop. This can happen when running
+            # paired Zernike estimation and the number of intra/extra stamps
+            # is not the same
+            if zk is None:
```
suberlak, line 244: So for one of the stamps not being there,  we get `None` rather than `[]` ? I'm checking against https://github.com/lsst-ts/ts_wep/blob/677cdd660f5273024841fcc97bbe04300c2a618a/python/lsst/ts/wep/task/estimateZernikesBase.py#L191

suberlak (approved):
Thanks for the update! Small comments to help document improvement,  no changes needed.

tests/task/test_calcZernikesTieTaskCwfs.py
jfcrenshaw: This test was added to reproduce the error, then the fix was written so this test passed.

Before, if e.g. there were more donuts in extraStamps than in intraStamps, it would hit a `zk = None`. This is because it had passed the number zernikes available. This happens because we are using `zip_longest` instead of `zip`, because this enables functionality in CalcZernikesUnpaired (where there are only intra or extra stamps, but not both). Hitting a `zk = None` caused an error because below it calls `zk[j - 4]`, and `None` is not subscriptable.

Now, when it gets to `zk = None` it stops, because it knows it has iterated through every paired stamp.

python/lsst/ts/wep/task/calcZernikesTask.py
jfcrenshaw: Yes, simply because we are using `zip_longest` instead of `zip`. However, this only happens when running inside the paired task.

Interestingly, I think the method you link to there is a little outdated. It still functions totally fine, however its structure is leftover from before the paired and unpaired estimation methods became separate tasks. I think it could be simplified now. (although definitely not high priority since it works fine).

python/lsst/ts/wep/task/donutStampSelectorTask.py
jfcrenshaw: It happens when you append a new DonutStamp to DonutStamps (like I do in the unit test I added). The DonutStamps append/extend methods don't extend the metadata at all, so e.g. `donutStamps.metadata.getArray("ENTROPY")` will return an array shorter than `donutStamps`.

We should probably create a ticket that fixes these methods so that metadata is properly carried along, however the append/extend methods are never used in the tasks, so it's low priority.

PR #281, opened 2024-10-29
Enable CutOutDonutsScienceSensorTask to operate for a pair with same-sign focusZ.

jfcrenshaw:
`CutOutDonutsScienceSensorTask` was previously written to require a pair of exposures with focusZ of opposite signs (i.e., one positive and one negative). If both focusZ had the same sign, it worried you had passed two exposures on the same side of focus, and raised an exception.

Apparently this was not desired and caused problems on night of 20241028. This PR removes this check and simply assigns the lower focusZ to intrafocal

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-12.5.0:
+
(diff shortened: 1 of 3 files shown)


gmegh (approved):
LGTM

PR #284, opened 2024-10-29
tickets/DM-47251: Added maxSelect to DonutStampsSelector

jfcrenshaw:
This PR implements a `maxSelect` config in `DonutStampsSelectorTask` to limit the number of stamps selected for Zernike estimation. We can now raise the number of stamps cutout, and allow this to throttle the number of Zernike estimates. This should ensure we get the same number of stamps per detector.

### DIFF
--- a/doc/versionHistory.rst
+++ b/doc/versionHistory.rst
@@ -6,6 +6,14 @@
 Version History
 ##################
 
+.. _lsst.ts.wep-12.6.0:
+
+-------------
+12.6.0
+-------------
+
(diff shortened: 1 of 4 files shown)


jbkalmbach (approved):
The code looks good but it needs testing that verifies it runs properly in the pipeline (that specifying this parameter in the pipeline yaml leads to only that many sources coming out in the butler output).

jfcrenshaw:
I will add that test after I fix donut_viz

PR #288, opened 2024-11-07
Tickets/DM-47442: Require convergence for TIE Zernike estimates

jfcrenshaw:
This also fixes:
-  a bug with the history-loading util
- a bug with changing the donut binning parameter not being picked up in the task

### DIFF
--- a/python/lsst/ts/wep/task/estimateZernikesBase.py
+++ b/python/lsst/ts/wep/task/estimateZernikesBase.py
@@ -69,12 +69,6 @@ class EstimateZernikesBaseConfig(pexConfig.Config):
         + "to the Noll index. In this case, indices 0-3 are always set to zero, "
         + "because they are not estimated by our pipeline.",
     )
-    binning = pexConfig.Field(
-        dtype=int,
-        default=1,
-        doc="Binning factor to apply to the donut stamps before estimating "
-        + "Zernike coefficients. A value of 1 means no binning.",
-    )
     saveHistory = pexConfig.Field(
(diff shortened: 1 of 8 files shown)


python/lsst/ts/wep/task/estimateZernikesBase.py
```diff
@@ -69,12 +69,6 @@ class EstimateZernikesBaseConfig(pexConfig.Config):
         + "to the Noll index. In this case, indices 0-3 are always set to zero, "
         + "because they are not estimated by our pipeline.",
     )
-    binning = pexConfig.Field(
```
jbkalmbach, line 72: Why did this have to move out of the base task? It looks like you moved the same thing into the individual tasks?

jbkalmbach (approved):
Looks good. I merged the necessary change in donut_viz to main and tested the pipeline successfully.

python/lsst/ts/wep/task/estimateZernikesBase.py
jfcrenshaw: It's because the binning is currently implemented inside the individual algorithms, instead of at the WfEstimator level. So this needs to be in the individual task configs so that the config is passed onto the individual algorithms. Configs at the base task level are filtered out when constructing individual algorithms: https://github.com/lsst-ts/ts_wep/blob/10fbca0b7270260e3bdd23e43f4c06f9c836e990/python/lsst/ts/wep/task/estimateZernikesBase.py#L104-L112

PR #290, opened 2024-11-11
tickets/DM-47480: sparse Zernike estimation

jfcrenshaw:


pipelines/production/comCamSimRapidAnalysisPipeline.yaml
```diff
@@ -5,21 +5,24 @@ tasks:
     class: lsst.ts.wep.task.generateDonutDirectDetectTask.GenerateDonutDirectDetectTask
     config:
       donutSelector.useCustomMagLimit: True
-      donutSelector.sourceLimit: 5
+      donutSelector.sourceLimit: 20
```
jmeyers314, line 8: Didn't we already change these?  Was that just on this branch?

jmeyers314 (approved):
Thanks for this!  LGTM.

pipelines/production/comCamSimRapidAnalysisPipeline.yaml
jfcrenshaw: Ah this is because I copied the ComCam config over to the ComCamSim config to keep them the same. I guess we forgot to update this number when we made this change in ComCam.

PR #296, opened 2024-11-29
Update current defaults for WEP pipelines

jfcrenshaw:
Includes sparsity, binning in Danish

### DIFF
--- a/pipelines/production/comCamRapidAnalysisPipeline.yaml
+++ b/pipelines/production/comCamRapidAnalysisPipeline.yaml
@@ -8,7 +8,7 @@ tasks:
   calcZernikesTask:
     class: lsst.ts.wep.task.calcZernikesTask.CalcZernikesTask
     config:
-      estimateZernikes.nollIndices: [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22]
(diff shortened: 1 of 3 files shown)


pipelines/production/comCamRapidAnalysisPipeline.yaml
```diff
@@ -8,7 +8,7 @@ tasks:
   calcZernikesTask:
     class: lsst.ts.wep.task.calcZernikesTask.CalcZernikesTask
     config:
-      estimateZernikes.nollIndices: [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22]
+      estimateZernikes.nollIndices: [4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 18, 19, 20, 21, 22, 27, 28]
```
jmeyers314, line 11: Do we trust 18, 19 more for TIE than for danish?

jmeyers314 (approved):
LGTM.  One question.

pipelines/production/comCamRapidAnalysisPipeline.yaml
jfcrenshaw: Not necessarily, but my testing found that it was sometimes needed for TIE convergence. We're currently running with a max 50 iterations though, so we could probably remove them if you like

PR #299, opened 2024-12-04
Upgrades to forward modeling util

jfcrenshaw:
Just a few small upgrades to the forward modeling util that were useful for tests I was running.

### DIFF
--- a/python/lsst/ts/wep/utils/modelUtils.py
+++ b/python/lsst/ts/wep/utils/modelUtils.py
@@ -106,11 +106,15 @@ def forwardModelPair(
     miscenterIntra : tuple, optional
         The amount by which the intrafocal donut is miscentered. A tuple of
         (dx, dy) in pixels. If None, a random value between +/- 2 is used
-        for each. (the default is None)
+        for each. Note the random non-zero offsets are default to avoid
+        pixel aliasing effects when examining ensembles of wavefront
+        estimation errors. (the default is None)
     miscenterExtra : tuple, optional
         The amount by which the extrafocal donut is miscentered. A tuple of
         (dx, dy) in pixels. If None, a random value between +/- 2 is used
-        for each. (the default is None)
+        for each. Note the random non-zero offsets are default to avoid
+        pixel aliasing effects when examining ensembles of wavefront
+        estimation errors. (the default is None)
     blendOffsetsIntra : Iterable or None, optional
         The blend offsets of the intrafocal donut.
         (the default is None)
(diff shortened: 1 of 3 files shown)


python/lsst/ts/wep/utils/modelUtils.py
```diff
@@ -97,17 +103,37 @@ def forwardModelPair(
         only specified for the intra or extrafocal donut, both donuts use
         that angle. If neither is specified, the same random angle is used
         for both. (the default is None)
+    miscenterIntra : tuple, optional
```
jbkalmbach, line 106: Why is it useful to have the default not be 0? Is it worth the bit of confusion that it may cause when somebody may assume that not entering anything in this optional parameter will still create a non-zero value?

python/lsst/ts/wep/utils/modelUtils.py
```diff
@@ -178,25 +204,38 @@ def forwardModelPair(
...
     if seeing > 0:
         intraStamp.image = convolve(intraStamp.image, atmKernel, mode="same")
-    # Normalize the flux
-    intraStamp.image *= fluxIntra * (1 + nBlends) / intraStamp.image.sum()
-    # Poisson noise
+    # Pseudo-Poissonian noise
     intraStamp.image += background
     intraStamp.image = np.clip(intraStamp.image, 0, None)
-    intraStamp.image = rng.poisson(intraStamp.image).astype(float)
+    intraStamp.image += rng.normal(size=intraStamp.image.shape) * np.sqrt(
```
jbkalmbach, line 239: Just trying to understand why this change?

jbkalmbach (approved):
Looks good. Minor comments.

python/lsst/ts/wep/utils/modelUtils.py
jfcrenshaw: With the Poisson, numpy sometimes complains "lam value too large". It's supposedly an issue with the float type in the array. I tried casting to different float types but didn't get it working. I decided it wasn't worth spending more time on, so just changed it to pseudo-Poissonian, which doesn't hit that error ¯\_(ツ)_/¯

python/lsst/ts/wep/utils/modelUtils.py
jfcrenshaw: I added it because using perfectly centered donuts causes some asymmetry in errors for Zernikes that are azimuthal pairs. I assume it's just a pixel aliasing thing. Adding small, random dithers fixes that issue. This amount of miscentering doesn't cause problems for the TIE or Danish, and I don't think it's bad to assume that stamps aren't perfectly centered. I'd be fine to set the default to 0 though if you think that's preferable.

jbkalmbach: Since the end product is more robust I think it's good to leave it. Maybe add a version of this explanation somewhere in the docstring or code so we understand when we look at this later.

PR #300, opened 2024-12-05
Danish joint fit

jfcrenshaw:
Danish can now jointly fit a pair of donuts, rather than fitting individually and then averaging Zernikes. I think this is likely more robust, e.g. I think it is more likely to properly distinguish astigmatism from DCR. Futhermore, it appears to be about the same speed as the fit-individually-and-average method. I have, therefore, made the joint-fit the new default.

### DIFF
--- a/python/lsst/ts/wep/estimation/danish.py
+++ b/python/lsst/ts/wep/estimation/danish.py
@@ -281,7 +281,7 @@ class DanishAlgorithm(WfAlgorithm):
             result = dict(result)
 
             # Unpack the parameters
-            dx, dy, fwhm, *zkFit = result.x
+            dx, dy, fwhm, *zkFit = result["x"]
             zkFit = np.array(zkFit)
 
             # Add the starting zernikes back into the result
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/estimation/danish.py
```diff
@@ -248,7 +297,7 @@ def _estimateSingleZk(
                     dy,
                     fwhm,
                     zkFit,
-                    sky_level=backgroundStd,
+                    sky_level=backgroundStd**2,
```
jmeyers314, line 300: Good catch!

python/lsst/ts/wep/estimation/danish.py
```diff
@@ -283,6 +332,172 @@ def _estimateSingleZk(
...
+                "zkFit": zkFit.copy(),
+                "zkSum": zkSum.copy(),
+                "model": modelImages[0],
+                "GalSimFFTSizeError": galSimFFTSizeError,
+            }
+            hist[I2.defocalType.value] = {
+                "image": imgs[1].copy(),
+                "variance": skyLevels[1],
+                "nollIndices": nollIndices.copy(),
+                "zkStart": zkStartI1.copy(),
```
jmeyers314: `zkStartI2` ?

jmeyers314 (approved):
LGTM!  One small comment.  I guess you need for #299 to merge and rebase before merging this one?

python/lsst/ts/wep/estimation/danish.py
jfcrenshaw: nice catch!

PR #306, opened 2025-01-28
Tickets/DM-44529: fix generateDonutDirectDetect for null selection

suberlak:
This PR adds a fix to the direct detection task, whereby if sources are detected, but none selected via `donutSelector`, the task would produce an error failing entire pipeline. Now it will return a `warning` similar to the case of no donuts detected, alongside with an empty donut table.

### DIFF
--- a/python/lsst/ts/wep/task/cutOutDonutsBase.py
+++ b/python/lsst/ts/wep/task/cutOutDonutsBase.py
@@ -611,36 +611,38 @@ reducing the amount of donut mask dilation to {self.bkgDilationIter}"
             # Save centroid positions as str so we can store in header
             blendStrX = ""
             blendStrY = ""
-
-            for blend_cx, blend_cy in zip(
-                catalogMeta["blend_centroid_x"][idx],
-                catalogMeta["blend_centroid_y"][idx],
-            ):
-                blend_final_x = blend_cx + donutRow["xShift"]
-                blend_final_y = blend_cy + donutRow["yShift"]
-                blendStrX += f"{blend_final_x:.2f},"
-                blendStrY += f"{blend_final_y:.2f},"
-            # Remove comma from last entry
-            if len(blendStrX) > 0:
-                blendStrX = blendStrX[:-1]
-                blendStrY = blendStrY[:-1]
-            else:
-                blendStrX = None
-                blendStrY = None
-            finalBlendXList.append(blendStrX)
-            finalBlendYList.append(blendStrY)
+            if len(catalogMeta["blend_centroid_x"])>0:
+                for blend_cx, blend_cy in zip(
+                    catalogMeta["blend_centroid_x"][idx],
+                    catalogMeta["blend_centroid_y"][idx],
+                ):
+                    blend_final_x = blend_cx + donutRow["xShift"]
+                    blend_final_y = blend_cy + donutRow["yShift"]
+                    blendStrX += f"{blend_final_x:.2f},"
+                    blendStrY += f"{blend_final_y:.2f},"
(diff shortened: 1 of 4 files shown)


python/lsst/ts/wep/task/generateDonutDirectDetectTask.py
```diff
@@ -273,27 +300,26 @@ def run(self, exposure, camera):
...
-                [detectorName] * len(donutCatUpd), dtype=str
-            )
+            # If at least one donut got selected, update the column names
+            # and content
+            if len(donutCatSelected) > 0:
+                donutCatUpd = self.updateDonutCatalog(donutCatSelected, exposure)
+                donutCatUpd["detector"] = np.array(
+                    [detectorName] * len(donutCatUpd), dtype=str
+                )
+            # If no donuts got selected, issue a warning and return an enpy
```
jbkalmbach: typo "enpy"

tests/task/test_generateDonutDirectDetectTask.py
```diff
@@ -236,6 +269,39 @@ def testTaskRun(self):
...
+            expected_metakeys,
+        )
+
+        # Test the behavior when source selection is turned off
+        self.task.config.doDonutSelection = False
+        taskOut_S11_noSelection = self.task.run(
+            exposure_S11,
+            camera,
+        )
+        # Check that the expected columns are present
```
jbkalmbach, line 300: Can you add a test to make sure there are sources when you do this just to check that it's not an empty catalog?

jbkalmbach (approved):
Looks good. Minor comments. Thanks for fixing this!

python/lsst/ts/wep/task/cutOutDonutsBase.py
```diff
@@ -611,36 +611,38 @@ def cutOutStamps(self, exposure, donutCatalog, defocalType, cameraName):
...
 
             # Prepare blend centroid position information
-            if len(catalogMeta["blend_centroid_x"][idx]) > 0:
-                blendCentroidPositions = np.array(
-                    [
-                        catalogMeta["blend_centroid_x"][idx] + donutRow["xShift"],
-                        catalogMeta["blend_centroid_y"][idx] + donutRow["yShift"],
-                    ]
-                ).T
+            if len(catalogMeta["blend_centroid_x"]) > 0:
```
jbkalmbach: Can you combine the logic here so you don't have the same command under two nested else statements?

jbkalmbach: Actually a better solution to all of this would be to check before the `for` loop starts on line 583 and:
```
if "blend_centroid_x" not in catalogMeta.keys(): # Or whatever best way to check there's no blend data
    catalogMeta["blend_centroid_x"] = []*len(donutCatalog)
    catalogMeta["blend_centroid_y"] = []*len(donutCatalog)
```
that way you won't have to add nested logic inside each iteration of the `for` loop.

suberlak: Ok, I implemented a different way that's more straightforward. We make the output of null selection (`donutSelector` was on , but no sources selected) same as  no selection (`donutSelector` was off, so that property was not calculated):  `donutCatalog.meta['blend_centroid_x'] = ""` 
Then in `cutOutDonutsBase` we similarly only use the `blend_centroid_x` if it is longer than 0. 

Also, I noticed that `finalBlendXList` gets created and updated, but not persisted anywhere (not in the `DonutStamp` nor in `stampsMetadata`. Shall I add it to the metadata, or remove?

PR #309, opened 2025-02-01
Tickets/DM-48706: safeguard against edge donuts

suberlak:
Hitherto if `donutSelector` was off,  we would include in `donutTable` donuts that were potentially too close to the detector edge to be cutout at the `cutOutDonuts` stage. This solution avoids including these in any `donutTable`.

### DIFF
--- a/python/lsst/ts/wep/task/generateDonutCatalogOnlineTask.py
+++ b/python/lsst/ts/wep/task/generateDonutCatalogOnlineTask.py
@@ -107,7 +107,7 @@ class GenerateDonutCatalogOnlineTask(pipeBase.Task):
         # refObjLoader handles the interaction with the butler repository
         # needed to get the pieces of the reference catalogs we need.
         refObjLoader = self.getRefObjLoader(refCatalogs)
-        margin = self.config.edgeMargin
+        edgeMargin = self.config.edgeMargin
         filterName = self.config.filterName
 
         try:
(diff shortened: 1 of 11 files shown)


python/lsst/ts/wep/task/generateDonutCatalogOnlineTask.py
```diff
@@ -45,6 +45,7 @@ class GenerateDonutCatalogOnlineTaskConfig(pexConfig.Config):
     doDonutSelection = pexConfig.Field(
         doc="Whether or not to run donut selector.", dtype=bool, default=True
     )
+    margin = pexConfig.Field(doc="Size of margin in pixels", dtype=int, default=80)
```
jbkalmbach: Something like `edgeMargin` would be more descriptive.

jbkalmbach: Same for other instances of margin in the PR.

python/lsst/ts/wep/task/generateDonutDirectDetectTask.py
```diff
@@ -123,6 +123,9 @@ class GenerateDonutDirectDetectTaskConfig(
     doDonutSelection = pexConfig.Field(
         doc="Whether or not to run donut selector.", dtype=bool, default=True
     )
+    donutStampSize = pexConfig.Field(
```
jbkalmbach: Keeping this parameter consistent as `edgeMargin` or something else will help users when using different tasks.

jbkalmbach (commented):
Looks ready for some tests. Thanks, I think this is a straightforward but very helpful update.

gmegh (approved):
LGTM! Thanks

PR #313, opened 2025-02-25
DM-49169: Cache Zernike bases

jmeyers314:
Caches the Zernike (gradient) basis arrays.  Speed up is ~12s -> ~10s on my test case of 5 donuts.


jfcrenshaw (approved):
LGTM

PR #314, opened 2025-02-25
Tickets/dm 49141: Run CutoutCwfsTask on single detectors and add new reassign task.

jbkalmbach:


python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
```diff
@@ -39,9 +37,42 @@
...
+    )
+    donutCatalog = connectionTypes.Input(
+        doc="Donut Locations",
+        dimensions=(
+            "visit",
+            "detector",
+            "instrument",
+        ),
+        storageClass="AstropyQTable",
+        name="donutTable",
```
jmeyers314, line 57: Probably good to explicitly add `multiple=False` since that's the bit you're changing from the base class here.  Same with the donut stamps.

python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
```diff
@@ -76,146 +107,52 @@ def __init__(self, **kwargs):
...
-            intraCatalog,
-            DefocalType.Intra,
+        # Check which defocal type applies to exposure
+        detectorName = exposure.getDetector().getName()
+        defocalType = None
+        if detectorName in self.extraFocalNames:
+            defocalType = DefocalType.Extra
+        elif detectorName in self.intraFocalNames:
+            defocalType = DefocalType.Intra
+        elif defocalType is None:
```
jmeyers314: I think this line is clearer as just `else:`

python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
```diff
@@ -76,146 +107,52 @@ def __init__(self, **kwargs):
...
         )
 
         # If no donuts are in the donutCatalog for a set of exposures
         # then return the Zernike coefficients as nan.
-        if (len(donutStampsExtra) == 0) or (len(donutStampsIntra) == 0):
-            return pipeBase.Struct(
-                donutStampsExtra=DonutStamps([]),
-                donutStampsIntra=DonutStamps([]),
-            )
+        if len(donutStampsOut) == 0:
```
jmeyers314, line 141: For this section, I prefer

```python
        if len(donutStampsOut) == 0:
            donutStampsOut = DonutStamps([])
        key = "donutStampsExtra" if defocalType == DefocalType.Extra else "donutStampsIntra"
        return pipeBase.Struct(**{key: donutStampsOut})
```

but I'll leave that choice to you.

python/lsst/ts/wep/task/reassignCwfsCutoutsTask.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+        detectorIdArrExtra = np.array(
+            [exp.dataId["detector"] for exp in inputRefs.donutStampsExtraIn]
+        )
+        detectorIdArrOut = np.array(
+            [exp.dataId["detector"] for exp in outputRefs.donutStampsExtraOut]
+        )
+
+        # Find cwfs detectors in the list of detectors being processed
+        runExtraIds = list(set(detectorIdArrExtra).intersection(extraFocalIds))
```
jmeyers314: Slight preference for `runExtraIds = sorted(set(detectorIdArrExtra).intersection(extraFocalIds))` here, but up to you.

tests/testData/gen3TestRepo/gen3.sqlite3
jmeyers314, line 1: Is this required?

tests/testData/pipelineConfigs/testCalcZernikesCwfsSetupPipeline.yaml
```diff
@@ -36,8 +36,12 @@ tasks:
   cutOutDonutsCwfsTask:
```
jmeyers314: Is there a use case for running `cutOutDonutsCwfsTask` without reassigning now?  I'm wondering if we should make the output default to `donutStampsExtraCwfs` instead of reconfiguring that here.  But maybe that makes existing repos unhappy or something...

tests/testData/pipelineConfigs/testReassignCwfsCutoutsPipeline.yaml
jmeyers314, line 1: The only difference between this file and `testReassignCwfsCutoutsPipeline.yaml` is 

```diff
<       doVariance: True
---
>       doVariance: False
```

Is that important, or could we just use a single pipeline?

jmeyers314 (approved):
Looks good to me!  I went ahead and rebased for you.  I have a few comments and some suggestions, but mostly minor.

tests/testData/gen3TestRepo/gen3.sqlite3
jmeyers314: Or maybe I accidentally added it in my rebase.  Sorry if so.

jbkalmbach: Yeah, it was me. It's necessary to change the dataset type definitions for the associated `log`, `metadata` and `config` dataset types that go with `CutOutDonutsCwfsTask`.

python/lsst/ts/wep/task/cutOutDonutsCwfsTask.py
jbkalmbach: I think that's much nicer, thanks.

tests/testData/pipelineConfigs/testCalcZernikesCwfsSetupPipeline.yaml
jbkalmbach: No, that should be fine. And yes, I think that's a good change that makes things easier to understand.

tests/testData/pipelineConfigs/testReassignCwfsCutoutsPipeline.yaml
jbkalmbach: Don't need a separate one.

PR #317, opened 2025-03-13
Update tests to be compatible with Python 3.12.

jbkalmbach:


tests/task/test_calcZernikesUnpairedTask.py
```diff
@@ -174,7 +174,7 @@ def testTable(self):
                         [row[f"Z{i}"].to_value(u.micron) for i in range(4, 29)]
                     )
                     i += 1
-                self.assertFloatsAlmostEqual(zkRaw1, zkRaw2, rtol=1e-7, atol=0)
+                self.assertFloatsAlmostEqual(zkRaw1, zkRaw2, rtol=1e-6, atol=0)
```
suberlak, line 177: Why is this failing just because of switch to 3.12?

suberlak (approved):
Thanks for the update!

tests/task/test_calcZernikesUnpairedTask.py
suberlak: 10 times less comparison accuracy... I guess within 1 nm rather than 0.1 nm is good enough, but I wonder why this change is needed

jbkalmbach: That or maybe a different version of numpy or something. The differences were just over 1e-7 at less that 1.1e-7 so it was probably just a small difference in some underlying library somewhere.

PR #318, opened 2025-03-20
Fix hanging in Jenkins testing by easing convergence parameters in nu…

jbkalmbach:
…mpy optimization methods.


python/lsst/ts/wep/estimation/tie.py
```diff
@@ -594,6 +620,7 @@ def _expSolve(
         dIdz: np.ndarray,
         nollIndices: np.ndarray,
         instrument: Instrument,
+        optimize: bool = True,
```
jfcrenshaw: Do you need to pass "optimize" around like this? Why don't you just use `self.optimizeLinAlg` directly in the `np.einsum()` calls?

jfcrenshaw (approved):
LGTM. Just one question about passing around the `optimize` argument inside the TIE

python/lsst/ts/wep/estimation/tie.py
jbkalmbach: Ah, good point, I missed that while I was cleaning things up. We can totally just do that.

tests/task/test_calcZernikesDanishTaskCwfs.py
```diff
@@ -19,8 +19,17 @@
...
+# flake8: noqa
 import os
 
+import threadpoolctl
+
+# Set environment variables first
+os.environ["OPENBLAS_NUM_THREADS"] = "1"
+os.environ["MKL_NUM_THREADS"] = "1"
+os.environ["OMP_NUM_THREADS"] = "1"
+threadpoolctl.threadpool_limits(limits=1, user_api="blas")
```
teweitsai: This part is duplicated in these 5 changed `test_xxx.py` files. Can we abstract this to a new function to avoid the duplication?

jbkalmbach: Yep, done.

PR #319, opened 2025-04-08
DM-47439: add fitDonutRadiusTask

suberlak:


python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -0,0 +1,275 @@
...
+    ConfigClass = FitDonutRadiusTaskConfig
+    _DefaultName = "FitDonutRadius"
+    """
+    Run donut radius fitting.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    @timeMethod
```
jbkalmbach: It doesn't look like this is actually modifying any default behavior. Do you need it here?

python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -0,0 +1,275 @@
...
+__all__ = [
+    "FitDonutRadiusTaskConnections",
+    "FitDonutRadiusTaskConfig",
+    "FitDonutRadiusTask",
+]
+
+
+class FitDonutRadiusTaskConnections(
+    pipeBase.PipelineTaskConnections,
+    dimensions=("instrument", "visit"),
```
jbkalmbach: Looks like you can add detector here too.

python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -0,0 +1,275 @@
...
+):
+
+    donutStampsIntra = ct.Input(
+        doc="Intrafocal Donut Stamps",
+        dimensions=("visit", "detector", "instrument"),
+        storageClass="StampsBase",
+        name="donutStampsIntra",
+        multiple=True,
```
jbkalmbach: Shouldn't need multiple here or `deferGraphConstraint`. Input connections should be the same as `CalcZernikesTask`.

python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+
+        butlerQC.put(
+            donutRadiiTable,
+            outputRefs.donutRadiiTable,
+        )
+
+    @timeMethod
+    def run(
+        self, intraStampsList: typing.List, extraStampsList: typing.List
```
jbkalmbach: You should get rid of multiple above and this should use DonutStamps as input. If you don't you won't be able to run each detector in parallel.

python/lsst/ts/wep/task/fitDonutRadiusTask.py
```diff
@@ -0,0 +1,275 @@
...
+        visitArray = []
+        radiiArray = []
+        leftEdgeArray = []
+        rightEdgeArray = []
+        centX0Array = []
+        centY0Array = []
+        detNameArray = []
+        defocalDistArray = []
+        failedFlagArray = []
+        # looping over all detector pairs
```
jbkalmbach: Should rewrite this to only use data from one detector in the task at a time.

jbkalmbach (changes requested):
You should rewrite this task to run on individual detectors as `CalcZernikesTask` does. That way it won't require all the data from all detectors be bottlenecked into a single task but can distribute each detector as parallel quanta.

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: Yes, it was a placeholder, can be  removed.

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: class signature updated

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: updated to be similar as in `CalcZernikesTask`

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: Rewritten

python/lsst/ts/wep/task/fitDonutRadiusTask.py
suberlak: Updated

gmegh (approved):
LGTM!

jbkalmbach (approved):

PR #320, opened 2025-04-12
Adjust ROTPA values after DM-49838

jbkalmbach:


suberlak:
Thanks for the quick fix!

suberlak (approved):
I assume the main change is ROTPA in the headers  - as long as the unit tests pass, we're good.

PR #322, opened 2025-04-13
Change reassign task to run per pair instead of per visit

gmegh:


TallJimbo (approved):
Looks good to me (I pair-coded this with @gmegh).

suberlak (approved):
Looks good, moving `donutStamps` from base to child classes makes it more versatile.

PR #324, opened 2025-04-15
Tickets/DM-50196: fix source selector and cutout donuts Cwfs for empty catalogs

suberlak:


jbkalmbach (approved):
Looks good. Just need to update the version number in the versionHistory file.

PR #325, opened 2025-04-15
Reduce test data needed by Jenkins

jbkalmbach:


python/lsst/ts/wep/utils/testUtils.py
```diff
@@ -46,3 +48,31 @@ def enforce_single_threading():
     # Limit threadpool to 1 thread for BLAS and OpenMP
     threadpoolctl.threadpool_limits(limits=1, user_api="blas")
     threadpoolctl.threadpool_limits(limits=1, user_api="openmp")
+
+
+def unzip_test_data(repoDir):
```
jmeyers314: Do we even need to gunzip these?  I thought most FITS readers can open `.fits.gz` natively.

suberlak (approved):
Looks good! Since this adds `unzip_test_data` in `testUtils`,  it probably should have a unit test in `test_testUtils` (like `getAmpImagesFromDir` did in `test_ioUtils`).

python/lsst/ts/wep/utils/testUtils.py
suberlak: Good point! `astropy.io.fits` definitely does handle `.fits.gz` without unzipping

jbkalmbach: I think we need them unzipped for the butler to read info from the headers.

jbkalmbach: But let me try it out a little and see if we can avoid this.

jbkalmbach: Happy to report that I was wrong. I will check that all tests are running and if so then I'll remove this function.

jbkalmbach:
> Looks good! Since this adds `unzip_test_data` in `testUtils`, it probably should have a unit test in `test_testUtils` (like `getAmpImagesFromDir` did in `test_ioUtils`).

Removed `unzip_test_data`

